PR #1, opened 2018-06-29
ci: initial circleci config

darahayes:
Hey folks,

just a little gift for you guys: a sample CircleCI config also with release stuff in place.

It's a skeleton where the actual steps such as testing, linting, docker builds and releases have not been implemented. However this config should work out of the box and you can think of it as an API/Interface that can be implemented in any way you'd like. As an example, we are using [Standard](https://standardjs.com/) for linting, but you guys can use whatever you'd like as long as you wire it up to `npm run lint` then you'll get it for free in CI. Hope that makes sense :)

Please let me know what you think.

I also have a very similar PR open against the data-sync-server repo which you can see here: https://github.com/aerogear/data-sync-server/pull/3

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,7 @@ version: 2
       steps:
         - checkout
         - setup_remote_docker
+        - run: npm run release:validate
         - run: npm run docker:build:release
         - run: npm run docker:push:release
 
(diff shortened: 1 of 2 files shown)


darahayes:
Sweet! You will see failing CI stuff for now but do not worry about this. Additionally, don't worry about implementing the docker build and release steps. @david-martin is going to implement this on the server and the implementation should be almost identical!

PR #5, opened 2018-07-02
Docker setup

pb82:
add scripts to build and push the image to dockerhub

ping @cfoskin

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker
+      - run: npm install
       - run: npm run build
       - run: npm run docker:build
       - run: npm run docker:push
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -17,6 +17,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker
+      - run: npm run build
```
pb82: @darahayes not sure if this step makes sense or if it still has the built artifacts from the `test` stage (npm run build builds the react ui app).

darahayes: @pb82 each stage is run inside a fresh container. For example, you can see how `checkout` is being performed in each stage.

I will leave this up to you. This depends on how your docker image is built. If the docker image depends on the `dist` folder already existing then you need to run `npm run build` first. In fact you will also need to run `npm install` first.

pb82: @darahayes Ok. Then don't you also need to add `npm install` here: https://github.com/aerogear/data-sync-server/blob/master/.circleci/config.yml#L20 ?

PR #6, opened 2018-07-02
Adds eslint configuration and refactor project and files

josemigallas:
TODO:

- [x] Include all files into eslint check
- [x] Fix build


server.js
```diff
@@ -1,35 +1,31 @@
...
-    return res.sendStatus(404);
-});
+App.get("*", (_, res) => res.sendStatus(404));
 
 exports.run = function (callback) {
     server = App.listen(port, () => callback(App));
 };
 
-exports.stop = function() {
+exports.stop = function () {
```
cfoskin: @josemigallas  should this be () =>  also?

josemigallas: Yes thanks, I definitely left some files unreviewed.

.babelrc
```diff
@@ -1,4 +1,4 @@
 {
-  "presets": ["env", "react"],
+  "presets": ["env", "react", "es2015"],
```
pb82: what does `es2015` give us that is not already in `react`?

package.json
```diff
@@ -1,40 +1,52 @@
...
+    "license": "Apache-2.0",
+    "repository": "https://github.com/aerogear/data-sync-ui",
+    "scripts": {
+        "lint": "eslint server test ui webpack.config.js --fix",
+        "server": "webpack --mode development && nodemon --watch ui --exec babel-node server/index.js",
+        "build": "webpack --mode production",
+        "clean": "rm -rf dist public/js/bundle.js node_modules",
+        "start": "node ./index.js",
+        "test": "NODE_ENV=test mocha",
+        "docker:build": "docker build -t aerogear/data-sync-ui:latest .",
```
pb82: needs a rebase, this file hase changed

.babelrc
josemigallas: It used to fix some vscode error I think, but I'll remove it.

package.json
josemigallas: should be fixed now

package.json
```diff
@@ -3,11 +3,14 @@
...
   "description": "AeroGear Sync Server UI",
   "main": "index.js",
-  "license": "MIT",
+  "license": "Apache-2.0",
+  "repository": "https://github.com/aerogear/data-sync-ui",
   "scripts": {
-    "server": "webpack --mode development && nodemon --watch ui index.js",
+    "lint": "eslint server test ui webpack.config.js --fix",
+    "server": "webpack --mode development && nodemon --watch ui --exec babel-node server/index.js",
     "build": "webpack --mode production",
```
pb82: don't we have to take `babel-node` into account here? In the docker container we are not using babel, we need to provide something that can be started with pure node.

josemigallas: Yes that's another lost during rebase, it's fixed now.

PR #9, opened 2018-07-04
Aerogear 3576 postgres

cfoskin:
Postgres integration - includes the creation of the tables when the app starts and updating the resolver to retrieve from postgres instead of the hard coded values.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,6 @@
     "express": "^4.16.3",
     "express-graphql": "^0.6.12",
     "graphql": "^0.13.2",
-    "graphql-relay": "^0.5.5",
-    "graphql-sequelize": "^8.3.1",
     "graphql-tag": "^2.9.2",
     "patternfly-react": "^2.10.2",
     "pg": "^7.4.3",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "build": "webpack --mode production && babel server --out-dir dist/",
     "clean": "rm -rf dist public/js/bundle.js node_modules",
     "start": "node ./dist/index.js",
-    "test": "mocha --require babel-core/register --recursive",
+    "test": "NODE_ENV=test mocha --require babel-core/register --recursive",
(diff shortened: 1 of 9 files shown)


cfoskin:
@josemigallas I just need to resolve the lint errors

package.json
```diff
@@ -24,11 +24,15 @@
     "express": "^4.16.3",
     "express-graphql": "^0.6.12",
     "graphql": "^0.13.2",
+    "graphql-relay": "^0.5.5",
+    "graphql-sequelize": "^8.3.1",
```
pb82: i think we can remove those two

cfoskin:
@pb82 @josemigallas  Tests are failing due to sequelize relying on a running postgres

package.json
cfoskin: 👍

pb82:
@cfoskin Ok, i can take a look at that and make the tests work without a database.

server/models/dataSource.js
```diff
@@ -0,0 +1,8 @@
+module.exports = (sequelize, DataTypes) => {
```
josemigallas: Could we stick with ES6 syntax? For example `export default function getDataSource() { ... }`.

server/models/index.js
```diff
@@ -0,0 +1,34 @@
...
+    });
+} else {
+    db = new Sequelize(config.database, config.username, config.password, {
+        host: config.host,
+        dialect: "postgres"
+    });
+}
+
+const database = db;
+const dataSource = require("./dataSource")(database, Sequelize);
```
josemigallas: Not a fan of this syntax 😅 could we import `getDataSource` on the top instead?

server/server.js
```diff
@@ -6,6 +6,7 @@ import setupGraphQLServer from "./gql";
 
 const App = express();
 let server = null;
+const database = require("./models");
```
josemigallas: Let's use `import` instead.

PR #11, opened 2018-07-04
Pubsub

pb82:
Add a pubsub module using node-pg-pubsub (https://github.com/voxpelli/node-pg-pubsub). No tests at the moment because it would require a running Postgresql instance. We can use the exported `publish` function to publish messages for the sync server.

@aliok We need to agree on channel names and payload format.


aliok:
About the channel name and the payload:
- We don't really need any payload. This publish event is basically a *poke* which tells sync servers to go and fetch everything, rebuild the GraphQL schema and feed it to Apollo server. If in the future we need different events, we can use a payload like `{event:"SCHEMA_UPDATE"}`. But I can't think of a case right now.
- Current channel name is `aerogear-data-sync-schema`. Sounds good?

aliok:
@pb82 @darahayes in regard to the architecture discussion we've had yesterday:

We need to have a abstraction layer over the pubsub mechanism.
I see you already have that in the admin server, but it is a bit different than we have in the sync server. See https://github.com/aerogear/data-sync-server/pull/13

In sync server, we basically didn't hardcode anything like channel name or what pubsub impl to use.
Instead, we have a env var called `SCHEMA_LISTENER_CONFIG` and we pick the correct pubsub impl based on that. Default config is like this (subject to change):
```
{ 
  type: 'postgres',
  config: { 
    channel: 'aerogear-data-sync-schema',
    database: 'postgres',
    username: 'postgres',
    password: 'mysecretpassword',
    host: '127.0.0.1',
    port: '5432' 
  } 
}
```

Here is the relevant code: https://github.com/aerogear/data-sync-server/blob/774067a352666f28e77c6c60a9934c6dbaaea4ef/server/lib/schemaListeners/schemaListenerCreator.js#L6

* It loads all modules under `listeners` directory
* It reads the env var above and uses the listener with type that matches the type in the config.

This is the pattern we use in choosing the right resolver builders, data source parsers etc as well.

@pb82 Let me know what you think about this structure. I think it is a good idea to use the same pattern in both servers.

In whatever way, I think it is better to parametrize the channel name anyway.

pb82:
@aliok Yes i think it's a good Idea to have the pubsub configurable. I will change it to a similar architecture than yours. Only i will call it `pubsub` instead of `listeners` because the admin ui doesn't listen on anything, it only publishes.

Regarding the channel name. There are different resources that can get updated: Schema, Data Sources, Resolvers. The channel name you picked implies that the schema was updated. Instead, can we either:

1. Have different channels, like `aerogear-data-sync-data-sources`, `aerogear-data-sync-schema`, etc.

1. Use a non-specific name like `aerogear-data-sync` and add information about the updated resource as payload.

What do you think?

server/configNotifiers/notifiers/postgres.js
```diff
@@ -0,0 +1,47 @@
+import PGPubSub from "pg-pubsub";
+import { info } from "../../logger";
+
+/**
+ * A pubsub implementation that uses Postgres notify/listen feature and the
+ * pg-pubsub lib (https://github.com/voxpelli/node-pg-pubsub)
+ */
+export class PostgresNotifier {
```
darahayes: ❤️ Before your most recent commit, I was going to suggest using a class here. This is nice

cfoskin:
@pb82  Looks good

pb82:
@darahayes We have npm scripts for `lint` and `lint:fix` for that but i fofgot to run them. @josemigallas what do you think about the pre-commit hook to run them automatically?

PR #15, opened 2018-07-06
ui and server use same dev db config

pb82:
Use a docker compose file for the database and also use the same default credentials as the sync server (so that you can run both at the same time).


README.md
```diff
@@ -6,7 +6,7 @@ The UI for the [AeroGear Data Sync Server](https://github.com/aerogear/data-sync
 
 1. There is a dependency on having a running postgres server. To run postgres in a docker container use the following command:
  
-    `docker run --rm --name=postgres -p 5432:5432 -e POSTGRES_PASSWORD=mysecretpassword postgres`
+    `docker compose up`
```
cfoskin: I think this needs to be `docker-compose up`

PR #17, opened 2018-07-06
Aerogear 3608 add mutation endpoint

cfoskin:
JIRA: https://issues.jboss.org/browse/AEROGEAR-3608

Added all mutations & resolvers and some tests.

### DIFF
--- a/server/gql/schema.js
+++ b/server/gql/schema.js
@@ -9,6 +9,9 @@ const Schema = buildSchema(`
     type Query {
         dataSources: [DataSource]
     },
+    type Mutation {
+            createDataSource(id: Int!, name: String!, config: String!): DataSource
+    },    
     type DataSource {
         id: Int!       
         name: String!
@@ -22,6 +25,22 @@ function listDataSources() {
     return dataSource.findAll();
 }
 
-const root = { dataSources: listDataSources };
+function createDataSource(name, type, config) {
+    info("createDataSource request");
+    return dataSource.create({
+        name: name,
+        type: type,
+        config: JSON.stringify({config})
+    }).then((dataSource) => {
+          return info("Data Source created: ", dataSource.name);
+    }).catch((err) => {
+        return info("error creating data source: ", err);
+    });
+}
+
+const root = {
+    dataSources: listDataSources,
+    createDataSource: createDataSource('Test Data Source', 'Postgres', '{test : test}' )
+};
 
 export { Schema, root };
--- a/server/gql/schema.js
+++ b/server/gql/schema.js
@@ -4,14 +4,22 @@ import { dataSource } from "../models";
 
 const Schema = buildSchema(`
     enum DataSourceType {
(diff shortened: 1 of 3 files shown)


server/gql/schema.js
```diff
@@ -4,24 +4,71 @@ import { dataSource } from "../models";
...
     },
     type Query {
-        dataSources: [DataSource]
+        dataSources: [DataSource],
+        getOneDataSource(id: Int!): DataSource
     },
+    type Mutation {
+        createDataSource(name: String!, type: DataSourceType!,  config: String!): DataSource
+        deleteDataSource(id: Int!): DataSource
+        updateDataSource(id: Int!, name: String!, type: DataSourceType!,  config: String!): DataSource
```
pb82: extra whitespace after `DataSourceType!,`

test/server/tests.js
```diff
@@ -1,5 +1,15 @@
 import { agent } from "supertest";
+import { graphql } from "graphql";
 import { run, stop } from "../../server/server";
+import { Schema, root } from "../../server/gql/schema";
+import { queries } from "./queries";
+
+const assert = condition => {
```
pb82: This could take a second argument with a message to return in the error if the condition is not met

server/gql/schema.js
```diff
@@ -4,24 +4,71 @@ import { dataSource } from "../models";
...
+    info("getOneDataSource request");
+    return dataSource.findById(id);
+};
+
+const deleteDataSource = ({ id }) => {
+    info(`deleteDataSource request for id ${id}`);
+    return dataSource.findById(id)
+        .then(foundDataSource => {
+            if (!foundDataSource) {
+                return;
```
josemigallas: Maybe we should reject here?

server/gql/schema.js
```diff
@@ -4,24 +4,71 @@ import { dataSource } from "../models";
...
+const updateDataSource = ({ id, name, type, config }) => {
+    info("updateDataSource request");
+    return dataSource.findById(id).then(foundDataSource => foundDataSource.update({
+        name,
+        type,
+        config
+    }));
+};
+
+const createDataSource = ({ name, type, config }) => {
```
josemigallas: Is there any field validation?

pb82: Graphql takes care of that. If that function is called it means the mutations has a valid format according to the schema.

test/server/tests.js
```diff
@@ -1,5 +1,15 @@
...
+import { graphql } from "graphql";
 import { run, stop } from "../../server/server";
+import { Schema, root } from "../../server/gql/schema";
+import { queries } from "./queries";
+
+const assert = (condition, message) => {
+    if (!condition) {
+        return new Error(`${condition} was not truthy: ${message}`);
+    }
+    return // eslint-disable-line
```
pb82: Semicolon missing

PR #18, opened 2018-07-09
update health endpoint

pb82:
Update the health endpoint to use the same format as on the server. Also remove animations from the tabs.


server/health/index.js
```diff
@@ -1 +1,68 @@
-export * from "./healthchecks";
+import { database as db } from "../models";
+
+export function runHealthChecks() {
+    // Checks database connectivity by auth attempt
```
josemigallas: Let's remove redundant comments, function is clear enough

server/health/index.js
```diff
@@ -1 +1,68 @@
...
+    function databaseConnectivityCheck() {
+        return db.authenticate();
+    }
+
+    function handleResolve({ label, promise }) {
+        return { label, promise: promise.then(() => ({ [label]: true })) };
```
josemigallas: This line is confusing, why we need the first label? In `handleReject` we don't.

server/health/index.js
```diff
@@ -1 +1,68 @@
...
+     *   {check1: <check1 status>},
+     *   {check2: <check2 status>}
+     * ]
+     */
+    function run() {
+        // Add additional health checks to this array
+        const checks = [{
+            label: "Database Connectivity",
+            promise: databaseConnectivityCheck()
+        }].map(handleResolve).map(handleReject);
```
josemigallas: Defining handleResolve and handleReject after run for a clearer reading

server/health/index.js
pb82: We pass the label to the function to be able to output a name with the health check. `handleResolve` is mapped over first: We have to return label again from there to be able to use it in the following `handleReject`. We don't return the label from there because we no longer need it (resolve and reject use it already) and more importantly: we are required to procue a list of promises, not obejcts with a label and a promise.

server/health/index.js
pb82: eslint doesn't like that, i moved them above run instead.

PR #19, opened 2018-07-10
Aerogear 7524 filtering

cfoskin:
Jira: https://issues.jboss.org/browse/AEROGEAR-7524

This should filter the dataSources by name. will start to work after 3 letters inputted by the user.

It should still list all data sources when no filter is used.

To create some test data : 

```mutation create {
  createDataSource(name: "TestDataSource", type: Postgres, config: "config string: string") {
    id
    name
    type
    config
  }
}
```

### DIFF
--- a/ui/components/common/CommonToolbar.js
+++ b/ui/components/common/CommonToolbar.js
@@ -2,7 +2,7 @@ import React from "react";
 import { Toolbar, Button } from "patternfly-react";
 import { DebounceInput } from "react-debounce-input";
 
-const CommonToolbar = ({ buttons, update }) => {
+const CommonToolbar = ({ buttons, onFilter }) => {
     const elements = buttons.map(button => (
         <Button
             style={{ float: "right" }}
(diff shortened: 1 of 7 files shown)


ui/components/common/CommonToolbar.js
```diff
@@ -1,7 +1,8 @@
 import React from "react";
 import { Toolbar, Button } from "patternfly-react";
+import { DebounceInput } from "react-debounce-input";
 
-const CommonToolbar = ({ buttons }) => {
+const CommonToolbar = ({ buttons, update }) => {
```
pb82: `update` is a bit generic, can we name it `onFilter` instead?

ui/components/data-sources/DataSourcesContainer.js
```diff
@@ -10,7 +10,8 @@ class DataSourcesContainer extends Component {
         super(props);
         this.state = {
             showModal: false,
-            modalText: null
+            modalText: null,
+            filter: {}
```
pb82: Why an object and not just a string?

cfoskin: The <Query> was failing with an error due to expecting an object and was getting a string. Changing it to an object solved the problem.

cfoskin: The variables object could also have many variables so it expects an object containing these

cfoskin:
@pb82  ok to merge?

PR #21, opened 2018-07-10
Add the dialog to create new data sources to the UI

josemigallas:
JIRA: https://issues.jboss.org/browse/AEROGEAR-3609

Additions:

- Modal "add data source" has forms
- Forms have validation
- Forms are cleared when modal closes
- "Add" button is disabled if any validation fails
- Data source is created when clicking "Add" button
- An alert is shown within the modal if an error ocurrs
- New data sources are added to the data sources list

Disclaimer:
- Under components/forms there are many components that are useless at the moment. They have been added to provide a boilerplate for future data source types and their customized forms.


ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -0,0 +1,216 @@
...
+                                            </MenuItem>
+                                            {/* More Data Source Types to be added */}
+                                        </DropdownButton>
+                                    </InputGroup.Button>
+                                </InputGroup>
+                            </Col>
+                        </FormGroup>
+
+                        {/* Config */}
+                        <FormGroup controlId="config" validationState={validations.config}>
```
pb82: Is the idea here to let the user provide a JSON string with the config? How would that work for the in-memory data source, something like `{"collection": "my_test_collection"}`? Would it not make more sense to let the user input the collection name and then deal with converting it to JSON on the backend?

cfoskin:
@josemigallas Im not 100% on how everything is working yet so while it looks good, I'll leave it to @pb82 to approve

ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -0,0 +1,207 @@
...
+import { InMemoryForms } from "./forms";
+import { DataSourceType } from "../../graphql/types/DataSourceType";
+import CreateDataSource from "../../graphql/CreateDataSource.graphql";
+import GetDataSources from "../../graphql/GetDataSources.graphql";
+
+const INITIAL_STATE = {
+    name: "",
+    type: DataSourceType.InMemory,
+    err: "",
+    inMemoryValues: {
```
darahayes: 'InMemoryValues' would suggest that this object is directly related to the In Memory Data Source, but looking at the code below, it looks like this is being used as a generic config object that gets passed along regardless of the data source type.

@josemigallas Let's say we have a Postgres datasource type, would this InMemoryValues object also be used to pass the postgres config (connection details, etc.) back to the API?

josemigallas:
@pb82 @cfoskin PR updated with new info from backend guys: config is no longer asked to the user but a `timestampData` checkbox.

@cfoskin Could you perhaps verify that after creation, the db entry looks like [this](https://github.com/aerogear/data-sync-server/blob/997aec51bf6fbb672419a573655ccb6718a88844/sequelize/seeders/20180704100455-notes-example.js#L6-L12)?

ui/components/data-sources/AddDataSourceDialog.js
josemigallas: No, this one is passed only to `InMemoryForms`. When Postgres type is added, there will be a `postgresValues` for the `PostgresForms` component. Everything has to be in the container/parent component tho, as it has the control over the state :(

ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    createDataSource() {
+        const { name, type, inMemoryValues } = this.state;
+
+        const config = JSON.stringify({ options: inMemoryValues });
+
+        return this.props.mutate({
+            variables: { name, type, config },
+            refetchQueries: [{ query: GetDataSources }]
```
pb82: Doesn't this require passing the filter variable?

pb82: as discussed, will be fixed by @cfoskin

PR #22, opened 2018-07-10
fonts must not be under css

pb82:
fonts must be under public, otherwise patternfly will not be able to find them. Icons will not work.


josemigallas:
Hey @pb82, some of the fonts must be under css/fonts, that's how patternfly-react.css is defined, ie:
`url("./fonts/OpenSans-Regular-webfont.woff2")`

Which is way they were moved in [f88a0e](https://github.com/aerogear/data-sync-ui/pull/12/commits/f88a0ec2f3ec5391de1b9c5e4033763ef8d8370e) under css and later they were divided between `css/fonts` and `fonts`.

pb82:
@josemigallas hmm, i was having problems with the icons when they were located under `css`

josemigallas:
Yes, that is my point, some must be under ./fonts (like the icons), and others are required under ./css/fonts.

PR #25, opened 2018-07-11
Code editor

pb82:
Adds a 'code editor' component that is basically just a <textarea> that has line numbers and accepts tabs. While for the schema editor we can at some point consider using a real IDE like ace (https://ace.c9.io/) i think for the request and response inputs we want something like this. Let me know what you think. It us used like this:

```jsx
<CodeEditor initialHeight={480} onChange={text = { ... }} />
```

and it looks like this:

bildschirmfoto 2018-07-11 um 12 11 22


ui/components/common/CodeEditor.js
```diff
@@ -0,0 +1,110 @@
+import React, { Component, createRef } from "react";
+
+import style from "./codeEditor.css";
```
pb82: @cfoskin @josemigallas what do you think about using styles in this way? (Requires a slight modification to the webpack config).

cfoskin: @pb82 👍  no issues with it

PR #26, opened 2018-07-11
move fonts to correct location

pb82:
move fonts to correct location


josemigallas:
I can still see this:
> GET http://localhost:8080/css/fonts/fontawesome-webfont.woff2?v=4.7.0 404 (Not Found)

otherwise good

pb82:
@josemigallas better now?

PR #28, opened 2018-07-11
empty state for data sources

pb82:
Renders a patternfly empty state when there are no data sources. It looks like this:

bildschirmfoto 2018-07-11 um 15 45 24

### DIFF
--- a/ui/components/data-sources/DataSourcesContainer.js
+++ b/ui/components/data-sources/DataSourcesContainer.js
@@ -30,7 +30,7 @@ class DataSourcesContainer extends Component {
     }
 
     render() {
-        const { showModal } = this.state;
+        const { showModal, filter } = this.state;
         return (
             <div>
                 <AddDataSourceDialog
@@ -44,7 +44,12 @@ class DataSourcesContainer extends Component {
                     }}
                 />
                 <div>
-                    <DataSourcesList filter={this.state.filter} /> {/*eslint-disable-line*/}
+                    <DataSourcesList
+                        filter={filter}
+                        onCreate={() => {
+                            this.addDataSource();
+                        }}
+                    />
                 </div>
             </div>
         );
--- a/ui/components/data-sources/DataSourcesList.js
+++ b/ui/components/data-sources/DataSourcesList.js
@@ -22,7 +22,7 @@ const DataSourcesList = ({ filter, onCreate }) => (
             if (items.length > 0) {
                 return (<ListView>{items}</ListView>);
             }
-            return <EmptyList action={onCreate} />;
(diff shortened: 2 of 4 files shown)


ui/components/data-sources/DataSourcesContainer.js
```diff
@@ -44,7 +44,12 @@ class DataSourcesContainer extends Component {
                     }}
                 />
                 <div>
-                    <DataSourcesList filter={this.state.filter} /> {/*eslint-disable-line*/}
+                    <DataSourcesList
+                        filter={filter}
+                        onCreate={() => {
```
josemigallas: onCreate={this.addDataSource}

ui/components/data-sources/DataSourcesList.js
```diff
@@ -1,11 +1,48 @@
...
+    EmptyStateInfo,
+    EmptyStateHelp,
+    EmptyStateAction
+} from "patternfly-react";
 
 import GetDataSources from "../../graphql/GetDataSources.graphql";
 import { DataSourcesListItem } from "./DataSourcesListItem";
 
-const DataSourcesList = ({ filter }) => (
+const EmptyList = ({ action }) => (
```
josemigallas: I think this one should be in a different file, better keep components small for maintainability

ui/components/data-sources/DataSourcesList.js
```diff
@@ -1,11 +1,48 @@
...
+        </EmptyStateInfo>
+        <EmptyStateHelp>
+                For more information have a look at the
+            <a href="https://docs.aerogear.org">&nbsp;docs</a>
+        </EmptyStateHelp>
+        <EmptyStateAction>
+            <Button
+                bsStyle="primary"
+                bsSize="large"
+                onClick={() => {
```
josemigallas: onClick={action}
But, since this component is closely related with "adding a data source", the callback should be called in accordance.

ui/components/data-sources/DataSourcesList.js
```diff
@@ -18,11 +55,10 @@ const DataSourcesList = ({ filter }) => (
                 <DataSourcesListItem item={item} key={item.id} />
             ));
 
-            return (
-                <div>
-                    <ListView>{items}</ListView>
-                </div>
-            );
+            if (items.length > 0) {
+                return (<div><ListView>{items}</ListView></div>);
```
josemigallas: We can remove the div

ui/components/data-sources/DataSourcesContainer.js
pb82: Unfortunately it won't work then because it needs to be called in the context of DataSourcesContainer.

PR #29, opened 2018-07-11
Implement Edit Data Source

josemigallas:
JIRA: https://issues.jboss.org/browse/AEROGEAR-3544

This PR introduces **big** changes to the DataSourcesDialog:

- AddDataSourceDialog has been repurposed into an "abstract" `BaseDataSourceDialog`.
- `AddDataSourceDialog` and `EditDataSourceDialog` are new children classes that shares common validation logic.
- Each Component implement a different query (create and update).
- BaseDataSourceDialog has "abstract" methods that children classes must implement in order to correctly setup things like title, disabled forms, the submit button.

**Important** changes in `DataSourcesContainer`:

- A new state property has been created `selectedDataSource` that is updated whenever the "Edit" option under the three dots dropdown is selected. That property is then passed into the appropriate modal as a prop. This approach could (or would) be used to "Delete Data Source" as well. Pinging @cfoskin .

### DIFF
--- a/ui/components/data-sources/AddDataSourceDialog.js
+++ b/ui/components/data-sources/AddDataSourceDialog.js
@@ -8,14 +8,14 @@ import GetDataSources from "../../graphql/GetDataSources.graphql";
 const INITIAL_STATE = {
     name: "",
     type: DataSourceType.InMemory,
-    err: "",
-    inMemoryValues: {
+    options: {
         timestampData: true
     },
+    err: "",
     validations: {
         name: null,
         type: "success",
-        inMemoryValues: "success"
+        options: "success"
     }
 };
 
@@ -37,9 +37,9 @@ class AddDataSourceDialog extends BaseDataSourceDialog {
     }
 
     createDataSource() {
-        const { name, type, inMemoryValues } = this.state;
+        const { name, type, options } = this.state;
 
-        const config = JSON.stringify({ options: inMemoryValues });
+        const config = JSON.stringify({ options });
 
         return this.props.mutate({
             variables: { name, type, config },
--- a/ui/components/data-sources/AddDataSourceDialog.js
+++ b/ui/components/data-sources/AddDataSourceDialog.js
@@ -1,22 +1,6 @@
-import React, { Component } from "react";
-import {
-    Alert,
-    Modal,
-    Icon,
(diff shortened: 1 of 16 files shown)


ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -65,139 +47,16 @@ class AddDataSourceDialog extends Component {
...
-                return (
-                    <InMemoryForms
-                        values={inMemoryValues}
-                        onValuesChange={values => this.onInMemoryValuesChange(values)}
-                    />
-                );
-            default:
-                return null;
-        }
+    isDisabled() {
```
pb82: Can be a constant instead of a function if all it does is return false.

psturc:
Verified the functionality, just a little detail about the buttons naming:
* JIRA description says we should change the "Add" button to "Save"
* changed name of the button (Edit) is displayed also in the form when I'm **adding** a new data source

screen shot 2018-07-12 at 10 32 53

ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -1,203 +1,62 @@
...
-    err: "",
-    inMemoryValues: {
+    options: {
         timestampData: true
     },
+    err: "",
     validations: {
         name: null,
         type: "success",
-        inMemoryValues: "success"
```
josemigallas: @darahayes I think I understand what you really meant in last PR with this field. At that time I didn't realize these "values" really belonged to `config.options`

josemigallas: Sorry wrong line, I meant [line 28](https://github.com/aerogear/data-sync-ui/pull/29/files#diff-017239a63889a936cf36a053872fe929L28)

josemigallas:
@darahayes @psturc  @pb82 
Thank you so much for the reviews.. if only you could do a quick review it again 😅 

- [x] Fixed dialog button name
- [x] DataSource full editable (name and options/config)
- [x] Some minor refactorizations

ui/components/data-sources/AddDataSourceDialog.js
pb82: Ah thanks for the explanation ;-)

PR #31, opened 2018-07-12
Schema editor v2

pb82:
Schema Editor:

- [x] Edit Schema
- [x] Save Schema
- [x] Download Schema
- [x] Render Schema Structure
- [ ] Edit resolver mappings (will be done in another view)

screenshot_20180712_120153


psturc:
Functionality seems to be working fine, I'll leave code review to domain experts.
Question:
* should the filter appear also in Schema section?
* when some schema is inserted, saved and the page is refreshed, despite the fact that schema is still present in the textarea, after hitting **Save**, error is thrown `GraphQL error: Syntax Error: Unexpected <EOF>` - yeah this is just an edge test case, I know :) maybe there's some easy way how to fix it - if not, let's create a jira

pb82:
@psturc thanks, that bug is fixed now! Regarding the filter: it's not implemented yet but it could filter the types maybe.

cfoskin:
Code looks good. Haven't tested functionality as Pavel seems to have done that.

ui/components/schema/StructureView.js
```diff
@@ -0,0 +1,136 @@
...
+    "__Type",
+    "__TypeKind",
+    "__Field",
+    "__InputValue",
+    "__EnumValue",
+    "__Directive",
+    "__DirectiveLocation"
+];
+
+class StructureView extends Component {
```
josemigallas: No need to make class component.

ui/components/schema/StructureView.js
```diff
@@ -0,0 +1,136 @@
...
+                    </Col>
+                    <Col xs={6} md={6}>
+                        {key}
+                    </Col>
+                </Row>
+            );
+        });
+    }
+
+    renderType(type, index) {
```
josemigallas: IMO this is clear indication that we should extract another class from here. But it's all right at the moment, we can leave it like this.

ui/components/schema/EmptyStructureView.js
```diff
@@ -0,0 +1,19 @@
...
+    EmptyState,
+    EmptyStateIcon,
+    EmptyStateTitle
+} from "patternfly-react";
+
+import style from "./structureView.css";
+
+const EmptyStructureView = () => (
+    <EmptyState style={{ margin: "5px" }} className={style.reducePadding}>
+        <EmptyStateIcon />
```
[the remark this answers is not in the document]
pb82: It just doesn't look good without anything in the empty state. We can surely improve it, use a better icon etc.

PR #35, opened 2018-07-13
Aerogear 7572

josemigallas:
JIRA: https://issues.jboss.org/browse/AEROGEAR-7572

### Description
screen shot 2018-07-13 at 13 44 38

Adds a warning message when "In Memory" is selected.

### Additional

Also, I consider a bit tasteless to show green when the field is valid, too many colors:
screen shot 2018-07-13 at 13 44 38

screen shot 2018-07-13 at 13 44 50

So that I changed the validationState value from `"success"` to `"ok"` just to avoid this style to be applied. In the end, it doesn't matter what this value is, as long as it is no `null || undefined || "error"`. This way it's more minimalistic and neutral:

screen shot 2018-07-13 at 13 48 55

### DIFF
--- a/ui/components/data-sources/BaseDataSourceDialog.js
+++ b/ui/components/data-sources/BaseDataSourceDialog.js
@@ -173,9 +173,6 @@ class BaseDataSourceDialog extends Component {
                                             <MenuItem eventKey={DataSourceType.InMemory}>
                                                 {DataSourceType.InMemory}
                                             </MenuItem>
-                                            <MenuItem eventKey={DataSourceType.Postgres}>
-                                                {DataSourceType.Postgres}
-                                            </MenuItem>
                                             {/* More Data Source Types to be added */}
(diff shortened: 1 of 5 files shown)


ui/components/data-sources/BaseDataSourceDialog.js
```diff
@@ -34,7 +35,9 @@ class BaseDataSourceDialog extends Component {
     }
 
     onTypeChange(type) {
-        const typeValidation = type && type.length < 255 ? "success" : "error";
+        let typeValidation = DataSourceType[type] ? "ok" : "error";
+        typeValidation = type === DataSourceType.InMemory ? "warning" : typeValidation;
+        console.log(typeValidation);
```
pb82: We should remove this

josemigallas: Oops

PR #36, opened 2018-07-13
notify server for data source and schema changes

pb82:
Notify the sync server when data sources or schema change.

### DIFF
--- a/server/gql/schema.js
+++ b/server/gql/schema.js
@@ -43,7 +43,7 @@ const createDataSource = async ({ name, type, config }) => {
         config
     });
 
-    publish(DEFAULT_CHANNEL, {reload: "DataSource"});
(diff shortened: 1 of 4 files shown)


server/gql/schema.js
```diff
@@ -57,24 +61,32 @@ const getOneDataSource = ({ id }) => {
...
+
+    const foundDataSource = await dataSource.findById(id);
+    if (!foundDataSource) {
+        return null;
+    }
+
+    const destroyed = await foundDataSource.destroy({force: true});
+    info(`Data source with id ${destroyed.id} deleted`);
+
+    publish(DEFAULT_CHANNEL, {reload: "DataSource"});
```
pb82: @aliok @darahayes does that payload format work for you?

pb82:
@cfoskin this is hard to verify on its own. Best to test in combination with the server.

PR #37, opened 2018-07-13
Don't refresh the page after hitting enter in filter

psturc:


ui/components/data-sources/DataSourcesContainer.js
```diff
@@ -35,6 +35,12 @@ class DataSourcesContainer extends Component {
         this.setState({ filter: { name } });
     }
 
+    handleKeyPress(event) {
+        if (event.key === "Enter") {
```
pb82: Very minor, but maybe use `keyCode` instead of the string value.

psturc: https://stackoverflow.com/a/34143412

ui/components/data-sources/DataSourcesContainer.js
```diff
@@ -35,6 +35,12 @@ class DataSourcesContainer extends Component {
         this.setState({ filter: { name } });
     }
 
+    handleKeyPress(event) {
```
josemigallas: I think this belongs to `CommonToolbar`. We want the filter to _never_ refresh the browser so the container (or any component rendering the toolbar) should not worry about this.

PR #40, opened 2018-07-16
refetch filter query when data source is deleted

pb82:
Fixes the problem where the data sources list was not updated when a filter was applied and a data source was deleted. The problem seems to be the way Apollo caches queries: we need to refetch the query without a filter value separately.

### DIFF
--- a/ui/components/data-sources/DeleteDataSourceDialog.js
+++ b/ui/components/data-sources/DeleteDataSourceDialog.js
@@ -6,11 +6,15 @@ import { graphql } from "react-apollo";
 import DeleteDataSource from "../../graphql/DeleteDataSource.graphql";
 import GetDataSources from "../../graphql/GetDataSources.graphql";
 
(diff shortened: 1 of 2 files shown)


ui/components/data-sources/DeleteDataSourceDialog.js
```diff
@@ -6,11 +6,15 @@ import { graphql } from "react-apollo";
...
 
-const DeleteDataSourceDialog = ({ showModal, dataSource, mutate, onClose }) => {
+const DeleteDataSourceDialog = ({ showModal, dataSource, filter, mutate, onClose }) => {
     const removeOneDatasource = dataSourceId => {
+        const { name } = filter;
         mutate({
             variables: { dataSourceId },
             refetchQueries: [{
+                query: GetDataSources,
+                variables: { name }
```
josemigallas: Why not replacing the other one? What's that one for?

josemigallas: OK nevermind this question, I answered to myself.

PR #41, opened 2018-07-16
add backend and tests for resolvers

pb82:
The backend part for the resolver mappings.

### DIFF
--- a/server/gql/schema.js
+++ b/server/gql/schema.js
@@ -1,6 +1,6 @@
 const { buildSchema } = require("graphql");
 const { info, warn, error } = require("../logger");
-const { dataSource, database, supportsiLike, schema } = require("../models");
+const { dataSource, database, supportsiLike, schema, resolver } = require("../models");
 const { compileSchemaString, formatGraphqlErrors } = require("./helper");
 const { publish, DEFAULT_CHANNEL } = require("../configNotifiers/configNotifierCreator");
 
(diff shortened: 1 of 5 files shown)


server/gql/schema.js
```diff
@@ -19,19 +19,38 @@ const Schema = buildSchema(`
         deleteDataSource(id: Int!): DataSource
         updateDataSource(id: Int!, name: String!, type: DataSourceType!, config: String!): DataSource
         updateSchema(id: Int!, schema: String!): Schema
+        upsertResolver(
+            id: Int
```
josemigallas: What's this id?

server/gql/schema.js
```diff
@@ -19,19 +19,38 @@ const Schema = buildSchema(`
...
+            schemaId: Int!
+            dataSourceId: Int! 
+            type: String!,
+            field: String!
+            requestMapping: String!
+            responseMapping: String!
+        ): Resolver
     },  
-    type DataSource {
+    type DataSource {
```
josemigallas: Additional spaces

server/gql/schema.js
```diff
@@ -56,6 +75,37 @@ const listDataSources = ({ name }) => {
...
+    }
+});
+
+const upsertResolver = async ({
+    id, schemaId, dataSourceId, type, field,
+    requestMapping, responseMapping
+}) => {
+    const properties = {
+        GraphQLSchemaId: schemaId,
+        DataSourceId: dataSourceId,
```
josemigallas: Is `properties.DataSourceId` upper case on purpose?

server/gql/schema.js
pb82: The id of the resolver. If provided it triggers an update, otherwise an insert.

server/gql/schema.js
pb82: Yes, thats how Sequelize expects them.

pb82: Could be overridden i guess. There is also a different API that we can use to add a resolver to a schema and data source but that would require us to fetch those instances first.

server/gql/schema.js
josemigallas: Maybe a silly question but doesn't that usually depend on the datasource id?

pb82: No, Resolver and Data Source are two different things. They are associated (one to many: one data source cab be used by many resolvers) but different tables and GraphQl entities.

PR #44, opened 2018-07-17
Adds loading spinner

josemigallas:
JIRA: https://issues.jboss.org/browse/AEROGEAR-7575

Shows a spinner whenever Apollo is "loading". Can't change this prop using React extension so in order to verify simply hardcode: `if (loading) -> if(true)`.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -56,3 +56,7 @@ span.list-item-name {
 .nav-tabs > li > a:hover{
     border: 1px solid lightgrey;
 }
+
+.spinner {
+    margin-top: 100px;
+}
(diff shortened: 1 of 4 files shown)


josemigallas:
> Why the change of div to Fragment?

**TLDR**: to not to add additional nodes to the DOM.

When rendering multiple components, React requires them to be wrapped around some node. The problem is normally you're rendering these multiple components inside a parent component, like:
```
<Container>
    <SomeComponentWithMultipleElements>
</Container>
```
So the fix is just wrapping everything around divs, which adds additional nodes to the DOM that are not used. `React.Fragment` allows you to by-pass this requirement, without adding any additional nodes. (see [Fragments](https://reactjs.org/docs/fragments.html))

PR #47, opened 2018-07-17
Adds shortcut for save schema

josemigallas:
JIRA: https://issues.jboss.org/browse/AEROGEAR-7577

Changes:
- Pressing `ctrl + s` when editor is focused triggers #save method.
- Toolbar is now a class component to allow disable/enable of buttons (it doesn't update only if props change, it has to have a state)
- "Save" button in Toolbar is disabled once the schema is saved, and enabled when schema changes.


ui/components/schema/SchemaContainer.js
```diff
@@ -44,7 +46,18 @@ class SchemaContainer extends Component {
     }
 
     onSchemaChange(schema) {
-        this.setState({ schema });
+        this.setState({ schema, saved: false });
+    }
+
+    @keydown("ctrl + s")
```
pb82: Crazy stuff

aliok:
I was able to verify the change in the UI. Can't speak about the code.

@pb82 @josemigallas I think it is a good idea to disable the save button. This gives some feedback to the user that the schema is saved. But, can we show an additional message like "Schema Saved" or something? For example like http://www.patternfly.org/pattern-library/communication/toast-notifications/

Take this as a general comment. Similar lack of feedback exists in saving a data source etc.

pb82:
@aliok I think this is a reasonable suggestion. We can display a notification when the schema was saved succesfully and only enable the button if the document has changed.

aliok:
@josemigallas don't let my comment about feedback block this PR.
@pb82 shall I create a JIRA ticket for that?

aliok:
@josemigallas Does it also work with Cmd+S ? If not, should it? I am not on a Mac, can't test.

josemigallas:
@aliok 
> can we show an additional message like "Schema Saved" or something?

Definitely, thanks for creating the JIRA.

> Does it also work with Cmd+S ?

CMD+S won't work. In their docs they say it should be possible with`@keydown("crtl + s", "cmd + ")`, but it simply doesn't work. I even files [an issue](https://github.com/glortho/react-keydown/issues/86).

@pb82 @aliok What we should add IMO is an indication of the shortcut. I tried adding a tooltip to the button but they way we have implemented the Toolbar, it's simply too complex to add tooltips. Any ideas?

pb82:
@josemigallas Why is it too complex? A tooltip could be passed to the toolbar as an additional (optional) parameter per button.

josemigallas:
Again, is not about how to tell the toolbar but how to actually render the buttons. I will update the code to show you.

josemigallas:
@aliok @pb82 This what it would look like: https://github.com/josemigallas/data-sync-ui/commit/f45a01650e83710c6b0aad3545be1f38a1550631

I am going to merge this anyway first, we can add the tooltip later.

PR #48, opened 2018-07-18
fix: calculate height of schema editor

psturc:
https://issues.jboss.org/browse/AEROGEAR-7596

* didn't find CSS solution for this issue - combining viewport height with overflow property could solve this issue, but it messes up **data sources** tab
* main problem with JS solution is that the distance of flexwrapper element from top can't be calculated before the element is displayed - this was the reason to use `componentDidUpdate` lifecycle method
* **this is just a suggestion of how we could fix that - feel free to (especially code owners) commit to this branch to make the code more elegant or even reject this**

### DIFF
--- a/ui/components/schema/SchemaContainer.js
+++ b/ui/components/schema/SchemaContainer.js
@@ -19,22 +19,19 @@ class SchemaContainer extends Component {
     constructor(props) {
         super(props);
         this.state = INITIAL_STATE;
-        this.flexWrapper = null;
-
-        this.setflexWrapperRef = e => {
-            this.flexWrapper = e;
-        };
+        this.flexWrapper = React.createRef();
     }
 
     updateDimensions() {
         this.setState({
-            height: window.innerHeight - this.flexWrapper.getBoundingClientRect().top
+            height: window.innerHeight - this.flexWrapper.current.getBoundingClientRect().top
         });
     }
 
     componentDidUpdate() {
-        const currentHeight = this.flexWrapper.style.height;
-        const distanceFromTop = this.flexWrapper.getBoundingClientRect().top;
+        const flexWrapperElement = this.flexWrapper.current;
+        const currentHeight = flexWrapperElement.style.height;
+        const distanceFromTop = flexWrapperElement.getBoundingClientRect().top;
         if (currentHeight === "0px" && distanceFromTop !== 0) {
             this.updateDimensions();
         }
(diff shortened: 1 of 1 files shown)


ui/components/schema/SchemaContainer.js
```diff
@@ -19,16 +19,25 @@ class SchemaContainer extends Component {
     constructor(props) {
         super(props);
         this.state = INITIAL_STATE;
+        this.flexWrapper = null;
+
+        this.setflexWrapperRef = e => {
```
pb82: The new way to create refs in react is to use `React.createRef()`: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

ui/components/schema/SchemaContainer.js
```diff
@@ -19,16 +19,25 @@ class SchemaContainer extends Component {
...
+            this.flexWrapper = e;
+        };
     }
 
     updateDimensions() {
         this.setState({
-            height: window.innerHeight - this.calculateHeaderHeight()
+            height: window.innerHeight - this.flexWrapper.getBoundingClientRect().top
```
pb82: Using the `top` property of the container is a great Idea!

josemigallas:
Honestly and all due respect, I don't see any real changes with this approach, it's still doing all in code instead of CSS. I am taking a look at this now, reviewing all CSS.

pb82:
@josemigallas the advantage of Pavel's approach is that we no longer hardcode the header height.

aliok:
Can't comment about the code but I was able to verify this change locally.

I played around and resized the window. 
* The schema editor grows/shrinks correctly.
* No unnecessary scroll bar shown at the document level

psturc:
@josemigallas of course, I also think that CSS solution would be more suitable. I pushed this guy because I wanted to start a discussion around this and because I spent less time on it than on CSS with no solution at all - but I'm no CSS expert. 
In fact, I kinda hate it a lot haha.
@aliok  @pb82 thanks for verification!

ui/components/schema/SchemaContainer.js
```diff
@@ -19,16 +19,22 @@ class SchemaContainer extends Component {
...
         });
     }
 
-    componentWillMount() {
-        this.updateDimensions();
+    componentDidUpdate() {
+        const flexWrapperElement = this.flexWrapper.current;
+        const currentHeight = flexWrapperElement.style.height;
+        const distanceFromTop = flexWrapperElement.getBoundingClientRect().top;
+        if (currentHeight === "0px" && distanceFromTop !== 0) {
```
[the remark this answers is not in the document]
psturc: Without checking the initial state (`currentHeight === "0px"`) it would end up calling `this.updateDimensions()` indefinitely (because  each update would trigger `componentDidUpdate()` method and the condition `distanceFromTop !== 0` would be always true)

psturc:
Done in https://github.com/aerogear/data-sync-ui/pull/52

PR #51, opened 2018-07-18
Trim data source name

aliok:
2 things:

* Show validation error when user types a blank data source name (blank means : `"     "`)
* Trim the data source name when saving. This is done on the backend. Sync UI loads the data from the backend upon saving anyway.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -29,7 +29,14 @@
         "react/sort-comp": "off",
         "react/no-did-update-set-state": "off",
         "jsx-a11y/label-has-for": "off",
-        "no-underscore-dangle": "off"
+        "no-underscore-dangle": "off",
+        "no-param-reassign": [
+            "error",
+            {
+                "props":
+                false
+            }
+        ]
     },
     "plugins": ["mocha"],
     "env": {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -33,8 +33,7 @@
         "no-param-reassign": [
             "error",
             {
-                "props":
-                false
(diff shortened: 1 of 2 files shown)


pb82:
@aliok Thanks, looks good! Please run `npm run lint:fix` to fix the exlint errors.

aliok:
@pb82 
One of the lint errors was problematic.

There this rule in ESLint: https://eslint.org/docs/rules/no-param-reassign

It doesn't allow this:
```
function foo(bar) {
    bar = 13;
}
```

While this case makes sense, it also disallows changing any props of the passed parameter:
```
function foo(bar) {
    bar.prop = "value";
}
```

This means, you can't create functions that modify anything in the params. In this case however, I needed to modify the name prop. IMO, this rule is limiting things too much. Imutability of that extent is unnecessary IMO.

I changed it in the ESLint config file to allow changing param props.

The alternative would be cloning the object, modifying it and returning it.

Let me know what you think.

pb82:
@aliok just disable the rule in .eslintrc and don't tell @josemigallas :-P

pb82:
@aliok Hmm, what about something like:

```
function foo(bar) {
    { prop } = bar;
    const result = { prop: prop.trim() };
}
```

aliok:
@josemigallas what do you think?

```
function foo(bar) {
    { prop } = bar;
    const result = { prop: prop.trim() };
}
```
That's too advanced JS for me @pb82 :smile: 

Seriously: that's one of way of going with the cloning approach I mentioned earlier.

PR #52, opened 2018-07-18
Fixes Schema Editor Height with CSS

josemigallas:
JIRA: https://issues.jboss.org/browse/AEROGEAR-7596

This is an alternative for #48 using only CSS. Height is set with `calc(100vh - 206px)` where `206px` is the header and toolbar height.

**EDIT**: forced push and removed all changes not related with the CSS height solution.

~~Changes:~~
- ~~CodeEditor initial props: uses now es6 default value syntax.~~
- ~~CodeEditor lineHeight prop: removed and added to styles.~~
- ~~Inline styles has been moved to css file.~~
- ~~Style class names has been renamed with a react-friendly name.~~
- ~~All resize logic in container has been removed.~~
- ~~SchemaContainer HTML has been simplified, now it uses Col with 2 Row.~~

### DIFF
--- a/ui/components/common/CodeEditor.js
+++ b/ui/components/common/CodeEditor.js
@@ -97,7 +97,7 @@ const CodeEditor = class extends Component {
                             lineHeight: `${lineHeight}px`,
                             whiteSpace: "pre",
                             overflowWrap: "normal",
-                            overflowY: "scroll"
+                            overflowY: "auto"
                         }}
                         className={style["editor-area"]}
                         onKeyDown={ev => this.onEditorKeyDown(ev)}
--- a/ui/components/schema/SchemaContainer.js
+++ b/ui/components/schema/SchemaContainer.js
@@ -11,7 +11,6 @@ import UpdateSchema from "../../graphql/UpdateSchema.graphql";
 import style from "./schemaContainer.css";
 
 const INITIAL_STATE = {
-    height: "100%",
     schema: "",
     error: null,
     saved: true
@@ -22,24 +21,6 @@ class SchemaContainer extends Component {
     constructor(props) {
         super(props);
         this.state = INITIAL_STATE;
-    }
-
-    updateDimensions() {
-        this.setState({
-            height: window.innerHeight - this.calculateHeaderHeight()
-        });
-    }
-
-    componentWillMount() {
-        this.updateDimensions();
-    }
-
-    componentDidMount() {
-        window.addEventListener("resize", () => this.updateDimensions());
-    }
-
-    componentWillUnmount() {
-        window.removeEventListener("resize", () => this.updateDimensions());
     }
 
     calculateHeaderHeight() {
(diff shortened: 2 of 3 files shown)


aliok:
@josemigallas 
It works ...but, I see both horizontal and vertical scroll bars at the document level.

See this screenshot:

screenshot from 2018-07-18 14-40-57

josemigallas:
True, I think it might be a problem with some negative padding I added. I'll take a look.

ui/components/common/CodeEditor.js
```diff
@@ -79,28 +74,20 @@ const CodeEditor = class extends Component {
     }
 
     render() {
-        const { editorValue, lineHeight } = this.state;
+        const { editorValue } = this.state;
         return (
-            <div className={style["editor-container"]}>
+            <div className={style.editorContainer}>
```
cfoskin: @josemigallas  is this way of referencing css preferred to this? `{style["editor-container"]}`

ui/components/schema/SchemaContainer.js
```diff
@@ -105,17 +77,17 @@ class SchemaContainer extends Component {
         return (
             <React.Fragment>
                 <CommonToolbar buttons={this.getToolbarButtons()} />
-                <div ref={this.setflexWrapperRef} className={style.flexWrapper} style={{ height: this.state.height }}>
-                    <div className={style.left}>
+                <Row>
+                    <Col xs={8} style={{ paddingRight: "0px" }}>
```
cfoskin: Again I am just wondering about referencing css, is this way just because one small piece of css is all that is needed?

ui/components/common/CodeEditor.js
josemigallas: The only different is how to access the property (period or brackets).

cfoskin: 👍

pb82: @cfoskin Javascript names can't have dashes, so when your class has a dash in the name you need to use a string instead of the dot notation.

aliok:
Ok, no more horizontal bar. And, vertical bar only when necessary. Looks good

pb82:
@josemigallas I can now see two scrollbars:
bildschirmfoto 2018-07-19 um 15 36 49

aliok:
:( yeah, 2 scrollbars on Firefox. 1 scrollbar on Chrome

josemigallas:
I'm taking a look at it. It's not just the scrollbar, the problem is textarea's height is not filling the editor container, ending up with this:
screen shot 2018-07-19 at 17 04 09
The magic was really done by `display: inline-table;` which doesn't seem to work in Firefox.

josemigallas:
Please @aliok @pb82 @cfoskin @psturc have a look.

psturc:
@josemigallas works perfectly - tested in Firefox & Chrome.
Just one thing - should we have a comment/note in [css file](https://github.com/aerogear/data-sync-ui/pull/52/files#diff-5deba11d1666dd6baca4a05797872ef6R2) about the magic value "206px"?

josemigallas:
Very good point @psturc , comment added and now merging.

PR #54, opened 2018-07-19
fix editor tabs

pb82:
A fix for the editor tab problem. Verification steps:

1. Enter a valid schema and save it
1. Add some tabs and save again
1. Tabs should be preserved


josemigallas:
I think actually `this.onEditorChange()` would be more proper, since it propagates the change.

PR #56, opened 2018-07-20
give clear error message when trying to save an empty schema

pb82:
When a user tries to save an empty schema a clear error message is shown. Previously nothing happend (because the last valid one was sent).

Also added `ID` to the list of internal types that we filter from the structure view and remove the cross symbols from the field list: you can collapse the list by clicking the header already.


pb82:
@cfoskin Yes, thats the intended error.

PR #57, opened 2018-07-20
WIP Aerogear 7591

cfoskin:
WIP a start on implementing the architecture for the resolver dialog. Currently can show/hide it by setting the boolean in the code for` visible={showAddModal}`

Need to next somehow tie it to the buttons in the SchemaContainer.js

@pb82 @josemigallas  Does this seem ok so far for skeleton?

screen shot 2018-07-20 at 16 15 48

### DIFF
--- a/ui/components/schema/AddResolverDialog.js
+++ b/ui/components/schema/AddResolverDialog.js
@@ -34,7 +34,7 @@ class AddResolverDialog extends BaseResolverDialog {
     }
 
     createResolver() {
-        console.log("called create resolver");
+        console.log("called create resolver function");
     }
 
     clearForms() {
--- a/ui/components/schema/AddResolverDialog.js
+++ b/ui/components/schema/AddResolverDialog.js
@@ -2,12 +2,12 @@ import BaseResolverDialog from "./BaseResolverDialog";
 
 const INITIAL_STATE = {
     dataSourceName: "",
-    requestmapping: "",
+    requestMapping: "",
     responseMapping: "",
     err: "",
     validations: {
         name: null,
-        requestmapping: "warning",
+        requestMapping: "warning",
         responseMapping: "success"
     }
 };
--- a/ui/components/schema/AddResolverDialog.js
+++ b/ui/components/schema/AddResolverDialog.js
@@ -1,0 +1,54 @@
+import BaseResolverDialog from "./BaseResolverDialog";
+
+const INITIAL_STATE = {
+    dataSourceName: "",
+    requestmapping: "",
+    responseMapping: "",
+    err: "",
+    validations: {
+        name: null,
+        requestmapping: "warning",
+        responseMapping: "success"
+    }
+};
+
+class AddResolverDialog extends BaseResolverDialog {
+
+    constructor(props) {
+        super(props);
+        this.state = INITIAL_STATE;
+    }
+
+    getTitle() {
+        return "Add Resolver";
+    }
+
+    getSubmitTitle() {
+        return "Add";
+    }
+
+    onSubmit() {
+        this.createResolver()
+            .then(() => this.onClose())
+            .catch(({ message = "Error" }) => this.setState({ err: message }));
+    }
+
+    createResolver() {
+        console.log("called create resolver");
+    }
+
+    clearForms() {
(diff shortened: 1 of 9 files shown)


pb82:
@cfoskin Looks good so far! You can already fill the data sources dropdown by using the query we already have. Maybe it would make sense to already add the templates? The idea is that the users might not always write their query from scratch and we provide a number of templates like `Default In Memory Query` or `Default SQL Select with ID`. The templates could be filled from a static JSON file and we would add dropdowns over the request and response mapping editors to select one. Upon selection the valud in the editors would be set to the template value. Could also be very useful for testing: you just add your often used queries to the JSON file.

ui/components/schema/AddResolverDialog.js
```diff
@@ -0,0 +1,54 @@
+import BaseResolverDialog from "./BaseResolverDialog";
+
+const INITIAL_STATE = {
+    dataSourceName: "",
+    requestmapping: "",
+    responseMapping: "",
+    err: "",
+    validations: {
+        name: null,
+        requestmapping: "warning",
```
josemigallas: There's a typo in `requestmapping`, it should be camelCase.

ui/components/schema/AddResolverDialog.js
```diff
@@ -0,0 +1,54 @@
+import BaseResolverDialog from "./BaseResolverDialog";
+
+const INITIAL_STATE = {
+    dataSourceName: "",
+    requestmapping: "",
```
josemigallas: Here as well (see comment below).

ui/components/schema/BaseResolverDialog.js
```diff
@@ -0,0 +1,192 @@
...
+                                    </InputGroup.Button>
+                                </InputGroup>
+                            </Col>
+                        </FormGroup>
+
+                        {/* Resolver request mapping */}
+                        <FormGroup controlId="requestMapping" validationState={validations.requestMapping}>
+                            <Col sm={3}>Resolver</Col>
+                            <Col sm={9}>
+                                <div style={{
```
josemigallas: Better move this style into a baseResolverDialog.css

ui/components/schema/EditResolverDialog.js
```diff
@@ -0,0 +1,72 @@
...
+
+class EditResolverDialog extends BaseResolverDialog {
+
+    constructor(props) {
+        super(props);
+        this.state = INITIAL_STATE;
+    }
+
+    componentDidUpdate(prevProps) {
+        if (this.props.dataSource && this.props.dataSource !== prevProps.dataSource) {
```
camilamacedo86: IHMO it could be encapsulated in a function and called here in order to make it more readable. Following an example. 

```javascript
function boolean isPropertiesChanges(prevProps) {
    return this.props.dataSource && this.props.dataSource !== prevProps.dataSource
}

...
if (sPropertiesChanges(prevProps)) {
   {...}
}
```

ui/components/schema/AddResolverDialog.js
```diff
@@ -34,7 +34,7 @@ class AddResolverDialog extends BaseResolverDialog {
     }
 
     createResolver() {
-        console.log("called create resolver");
+        console.log("called create resolver function");
```
camilamacedo86: Could we not use a logger deb/lib here?

camilamacedo86: How can a function do 2 actions create and resolver? IMHO each function should do one action(encapsulation). 
Then here we will have something as follows. 

```javascript

create() {
   console.log("called create function");
}

resolver() {
   console.log("called resolver function");
}

```

Also, what exactly means resolver? Could we find a better word to clarify what should be done here?

cfoskin:
Hi @camilamacedo86  This is still a WIP as mentioned in the title, thanks for your reviews

camilamacedo86:
Hi @cfoskin,

It is fine. Your implementations is great 🥇 . I hope that the reviews are helpful. IHMO we should try to start the things as better/perfect as possible becuase then will be easer keep mantained and improve it. I am beliver of the "Broken Windows" theory. :P

cfoskin:
Closing this as no longer taking the modal approach

PR #59, opened 2018-07-23
Adds Feedback for Save Schema Button

josemigallas:
JIRA: https://issues.jboss.org/browse/AEROGEAR-7601

**SchemaContainer**'s state has 2 fields now:

- `saved`
- `saving`

Now since this action is asynchronous, button will be disabled and saying _Saving Schema..._ until the mutation has finished. If it's been saved successfully, button will state _Schema saved_, otherwise an error will pop up and the button will go back to _Save Schema_.

Also, added a new prop to **CodeEditor**: `disabled`, making the `textarea` disabled and dim.

Verification steps:

1. Wrap the mutation call `this.props.mutate(...)` inside `#save` with `setTimeOut` so that the action takes some time.
```javascript
save() {
    /* ... */
    setTimeout(() => {
        this.props.mutate({
            /* ... */
        });
    }, 1500);
}
```

2. Add a valid schema at tab *Schema* and click on save.
```
enum DataSourceType {
    InMemory,
    Postgres,
},
type Query {
    dataSources(name: String): String
}
```

3. Add some error to the schema, an error message should appear.
```
dataSources(name: String): void
```

4. Fix it and save again, the error should disappear.

### DIFF
--- a/ui/components/common/CodeEditor.js
+++ b/ui/components/common/CodeEditor.js
@@ -83,6 +83,7 @@ const CodeEditor = class extends Component {
     }
 
     render() {
+        const { disabled } = this.props;
         const { editorValue, lineHeight } = this.state;
         return (
             <div className={style["editor-container"]}>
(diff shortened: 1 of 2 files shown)


pb82:
👀

josemigallas:
I think it would, I'll push the changes.

PR #65, opened 2018-07-25
Postgres ds

pb82:
Add the postgres option to the data sources dialog. Also adds a nicer way to do validations.

### DIFF
--- a/ui/components/common/Validators.js
+++ b/ui/components/common/Validators.js
@@ -1,0 +1,35 @@
+const Validators = {
+    String: {
+        defined: s => s
+                && typeof s === typeof ""
+                && s.length > 0
+    },
+    Number: {
+        natural: n => !Number.isNaN(n) && parseInt(n, 10) > 0
+    }
+};
+
+/**
+ * Run a number of validations that have to be passed in an array of the form:
+ * [
+ *  <Validation Function 1>, <Validation Subject 1>
+ *  ...
+ *  <Validation Function N>, <Validation Subject N>
+ * ]
+ * @param validations The validations array
+ * @returns {string} "success" on success, otherwise "error"
+ */
+const Validate = validations => {
+    let result = false;
+
+    if (validations && validations.length % 2 === 0) {
+        result = true; // Init value
+        for (let i = 0; i < validations.length; i += 2) {
+            result = result && validations[i](validations[i + 1]);
+        }
+    }
+
+    return result ? "success" : "error";
+};
+
+export { Validators, Validate };
--- a/ui/components/common/Validators.js
+++ b/ui/components/common/Validators.js
@@ -1,11 +1,17 @@
 const Validators = {
     String: {
-        defined: s => s
+        nonBlank: s => s
                 && typeof s === typeof ""
-                && s.length > 0
+                && s.length > 0,
+        defined: s => s && typeof s === typeof "",
+        minLength: l => s => Validators.String.nonBlank(s) && s.length >= l,
+        maxLength: l => s => Validators.String.nonBlank(s) && s.length < l
     },
     Number: {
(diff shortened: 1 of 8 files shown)


ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -9,7 +9,16 @@ const INITIAL_STATE = {
     name: "",
     type: DataSourceType.InMemory,
     options: {
-        timestampData: true
+        // InMemory Options
+        timestampData: true,
+
+        // Postgres Options
+        url: "",
+        port: 5432,
```
cfoskin: Could it ever happen that this would be configured to be something different by the user?

ui/components/common/Validators.js
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Run a number of validations that have to be passed in an array of the form:
+ * [
+ *  <Validation Function 1>, <Validation Subject 1>
+ *  ...
+ *  <Validation Function N>, <Validation Subject N>
+ * ]
+ * @param validations The validations array
+ * @returns {string} "success" on success, otherwise "error"
+ */
```
camilamacedo86: 👍 Great comment !!!

pb82: Sure, i can add a real example

camilamacedo86: @pb82 it was merged without a better explanation.

ui/components/common/Validators.js
```diff
@@ -0,0 +1,43 @@
...
+                && typeof s === typeof ""
+                && s.length > 0,
+        minLength: l => s => Validators.String.nonBlank(s) && s.length >= l,
+        maxLength: l => s => Validators.String.nonBlank(s) && s.length < l
+    },
+    Number: {
+        natural: n => !Number.isNaN(n) && parseInt(n, 10) > 0
+    },
+    Port: {
+        valid: p => Validators.Number.natural(p) && p < 65536
```
camilamacedo86: Should it not be a parameter?

ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -8,9 +8,16 @@ import GetDataSources from "../../graphql/GetDataSources.graphql";
 const INITIAL_STATE = {
     name: "",
     type: DataSourceType.InMemory,
-    options: {
+    inMemoryOptions: {
         timestampData: true
     },
+    postgresOptions: {
+        url: "",
+        port: 5432,
```
camilamacedo86: Should it not be a parameter?

pb82: It is, thats just the default value (Postgres default port)

ui/components/common/Validators.js
pb82: Ports can't be >65536, thats a constant

ui/components/common/Validators.js
pb82: I added an example of how to use the validation function. It's not really general and only used for the dialogs in the UI.

ui/components/data-sources/AddDataSourceDialog.js
camilamacedo86: Yep, but it could be changed. Also, it should be better if was defined by a const do you don't think? Something as follows. 

```javascript
const POSTGRES_PORT = process.env.POSTGRES_PORT ; // it should came from a property setup
{....}
port: POSTGRES_PORT,
```

ui/components/common/Validators.js
camilamacedo86: So, should it not be better in a const var to make it clear? Following an example.

```javascript
const MAX_POSTHRES_PORT = 65536

{...}

valid: p => Validators.Number.natural(p) && p < MAX_POSTHRES_PORT
```

PR #67, opened 2018-07-25
better explanation for the validators

pb82:
Not really sure what else to add here.

### DIFF
--- a/ui/components/common/Validators.js
+++ b/ui/components/common/Validators.js
@@ -24,6 +24,10 @@ const Validators = {
  *  ...
  *  <Validation Function N>, <Validation Subject N>
  * ]
+ *
+ * This can be used to perform a number of validations against the available
+ * validators. Used to check dialog input values for correctness.
+ *
  * @param validations The validations array
  * @returns {string} "success" on success, otherwise "error"
  */


ui/components/common/Validators.js
```diff
@@ -24,6 +24,10 @@ const Validators = {
  *  ...
  *  <Validation Function N>, <Validation Subject N>
  * ]
+ *
+ * This can be used to perform a number of validations against the available
+ * validators. Used to check dialog input values for correctness.
+ *
```
camilamacedo86: Why is a par length true? What and why need to be check done when the length is an odd number?

pb82: It's always one validator per input value. So the total length of the array % 2 needs to be zero.

pb82: So if you want to validate 4 input values you would have 4 validator functions + 4 input values. Array length needs to be 8.

josemigallas: I find this function confusing, why would you even pass the names of the forms to validate if in the end you just want the reduced validation result?
Also, if validations is a 10 length array and the first one fails, it still iterate over the rest.
Finally, if the input is wrong (say a length 3 array) it returns a false negative.

josemigallas: Sorry I know it's late to say this but I didn't know this even existed.

pb82: You don't pass the name. You only pass the validation function and the value. I'll change it to exit early when the result is false.

pb82: @camilamacedo86 @josemigallas https://github.com/aerogear/data-sync-ui/pull/70

PR #68, opened 2018-07-26
task(improve): Improvements ref fixed port values in the code

camilamacedo86:
# Motivation:

* Make the propose of the number/port clear
* Create constants for const/fixed values

### DIFF
--- a/ui/components/common/Validators.js
+++ b/ui/components/common/Validators.js
@@ -1,3 +1,4 @@
+const MAX_VALID_PORT = 65536;
 const Validators = {
     String: {
         nonBlank: s => s
@@ -10,7 +11,7 @@ const Validators = {
         natural: n => !Number.isNaN(n) && parseInt(n, 10) > 0
     },
     Port: {
-        valid: p => Validators.Number.natural(p) && p < 65536
+        valid: p => Validators.Number.natural(p) && p < MAX_VALID_PORT
     },
     Boolean: {
         valid: b => typeof b === typeof true


ui/components/common/Validators.js
```diff
@@ -1,3 +1,4 @@
+const MAX_POSTGRES_PORT = 65536;
```
pb82: It's not the maximum value for Postgres, it't the maximum value any port can have.

ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -13,7 +14,7 @@ const INITIAL_STATE = {
     },
     postgresOptions: {
         url: "",
-        port: 5432,
+        port: serverConfig.postgresConfig.port,
```
pb82: We can't use `serverConfig.postgresConfig.port` here because those are two different databases. The one configured in `server/config.js` is the admin database while the one in the `AddDataSourceDialog` is a data source of type Postgres that can point to any database the user wants. The default value of 5432 is there because it's the default Postgres port (but the user can put in whatever they want).

pb82: Yes, this one can be changed.

ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -13,7 +14,7 @@ const INITIAL_STATE = {
     },
     postgresOptions: {
         url: "",
-        port: 5432,
+        port: DEFAULT_POSTGRES_PORT,
```
josemigallas: I see your point here but there's really no need for extracting _yet another_ constant for this value. `INITIAL_STATE` is already the extracted constant defining all default values.

camilamacedo86: 👍 Perfect!

pb82:
Closing because the changes are now in https://github.com/aerogear/data-sync-ui/pull/70

PR #70, opened 2018-07-26
early exit

pb82:
Early exit condition for the validation function. Also improves the comment.

### DIFF
--- a/ui/components/common/Validators.js
+++ b/ui/components/common/Validators.js
@@ -1,3 +1,4 @@
+const MAX_PORT_NUM = 65535;
 const Validators = {
     String: {
         nonBlank: s => s
@@ -10,7 +11,7 @@ const Validators = {
         natural: n => !Number.isNaN(n) && parseInt(n, 10) > 0
     },
     Port: {
-        valid: p => Validators.Number.natural(p) && p < 65536
+        valid: p => Validators.Number.natural(p) && p < MAX_PORT_NUM
     },
     Boolean: {
         valid: b => typeof b === typeof true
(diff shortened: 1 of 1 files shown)


ui/components/common/Validators.js
```diff
@@ -1,3 +1,4 @@
+const MAX_PORT_NUM = 65535;
```
josemigallas: Isn't this #68 's ? :P

pb82: Yes but i thought why not just do it all in one

pb82: @camilamacedo86 is that ok for you?

camilamacedo86: 👍perfect..  Just close that one.

camilamacedo86: @pb82, It is fine .. However, maybe something, as follows, could make clear what it is doing. 

```javascript 

/**
 * Run a number of validations that have to be passed in an array of the form:
 * [
 *  <Validator 1>, <Input Value 1>
 *  ...
 *  <Validator N>, <Input Value N>
 * ]
 * @param validations The validations array
 * @returns {string} "success" on success, otherwise "error"
 */
const Validate = validations => {
    // It always has to be exactly one validator per input value, so the total length
    // of the array mod 2 must be zero.
    function hasQtExpectedPositions() {
        return validations && validations.length % 2 !== 0;
    }
    function hasInputValue(arr, index) {
        return arr[index](arr[index + 1]);
    }
    // For each validator is expected an input value
    function hasExpectedValues(arr) {
        for (let i = 0; i < arr.length; i += 2) {
            if (!hasInputValue(arr, i)) {
                return false;
            }
        }
        return true;
    }
    if (!hasQtExpectedPositions() || !hasExpectedValues(validations)) {
        return "error";
    }
    return "success";
};

export { Validators, Validate };


```

pb82: @camilamacedo86 maybe this example makes it clear how it is used: https://github.com/aerogear/data-sync-ui/blob/master/ui/components/data-sources/BaseDataSourceDialog.js#L66

PR #73, opened 2018-07-26
added pino logging

cfoskin:
## Motivation
We want to have the logging in JSON format

## What
Use the pino logging library which is the same as the data-sync server and outputs everything in JSON format

## Why
So we can consume this later

## How
Change Winston for Pino

## Verification Steps
run the server : 

```
{"level":40,"time":1532622330134,"msg":"notifier unknown or disabled: postgres","pid":16137,"hostname":"cfoskin.local","v":1}
{"level":30,"time":1532622330204,"msg":"GraphQl server mounted at /graphql","pid":16137,"hostname":"cfoskin.local","v":1}
{"level":30,"time":1532622330269,"msg":"AeroGear Sync UI running on port 8080","pid":16137,"hostname":"cfoskin.local","v":1}
```

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


package.json
```diff
@@ -25,12 +25,14 @@
...
     "express": "^4.16.3",
     "express-graphql": "^0.6.12",
+    "express-pino-logger": "^3.0.2",
     "graphql": "^0.13.2",
     "graphql-tag": "^2.9.2",
     "lodash.some": "^4.6.0",
     "patternfly-react": "^2.10.2",
     "pg": "^7.4.3",
     "pg-pubsub": "^0.4.0",
+    "pino": "^4.17.5",
```
pb82: If we use pino we can remove the dependency to Winston.

pb82:
@cfoskin Looks good overall, can we replace the sql logging function [here](https://github.com/aerogear/data-sync-ui/blob/master/server/models/database.js#L17) with something like:

```
 logging: postgresConfig.logSql ? pino.debug : false
```

package.json
```diff
@@ -25,21 +25,22 @@
     "css-loader": "^1.0.0",
     "express": "^4.16.3",
     "express-graphql": "^0.6.12",
+    "express-pino-logger": "^3.0.2",
```
pb82: Does `express-pino-logger` already log any http requests? If so we can probably remove logs like `list data sources request`

cfoskin: Yes it does Peter, I will remove all cases in the resolvers where we do this

package.json
```diff
@@ -40,8 +40,7 @@
     "react-keydown": "^1.9.7",
     "sequelize": "^4.38.0",
     "sqlite3": "^4.0.1",
-    "style-loader": "^0.21.0",
-    "winston": "^3.0.0"
+    "style-loader": "^0.21.0"
```
camilamacedo86: the `ˆ` can install a major version with break changes which can impact the code.

PR #74, opened 2018-07-26
Reduce structure view

pb82:
## Motivation

Reduce the structure view in the schema editor to what we discussed in the UX review: no interactivity, only list the types and their properties. I suggest that we write a different component for the structure view in the resolver tab because it has a different organization and more features.

## Verification Steps

Paste this Schema and verify the types are rendered correctly:

```
  type Meme {
    id: ID! @isUnique
    photoUrl: String!
  }
  
  type Query {
    allMemes:[Meme!]!
  }
  
  type Mutation {
    createMeme(photoUrl: String!):Meme!
  }
  
  type Subscription {
    _: Boolean
  }
```

## Checklist:

## Progress

- [x] Finished task

## Screenshot

bildschirmfoto 2018-07-26 um 18 36 46


ui/components/common/GraphQLFormatters.js
```diff
@@ -0,0 +1,20 @@
+const formatType = type => {
+    let result = "";
+    switch (type.kind) {
+        case "LIST":
+            result += "[";
```
josemigallas: Let's use template strings instead.

ui/components/common/GraphQLFormatters.js
```diff
@@ -0,0 +1,20 @@
+const formatType = type => {
```
josemigallas: Not a component, should be located out of this folder. (same goes for validators but.. too late)

ui/components/schema/TypeList.js
```diff
@@ -133,26 +66,7 @@ class TypeList extends Component {
...
-                        return this.renderLoading();
-                    }
-                    if (error) {
-                        return this.renderError(error);
-                    }
-                    return this.renderList(data);
-                }}
-            </Query>
-        );
+        return this.renderList();
```
josemigallas: What is this method for?

pb82: Good question, i'll remove the call

PR #76, opened 2018-07-30
Sync models with data-sync-server

psturc:
## Motivation
https://github.com/aerogearcatalog/data-sync-apb/pull/10 (APB provision was failing)

## What
* added `server/models/subscription.js`
* added field `publish` to `server/models/resolver.js`

### DIFF
--- a/server/models/subscription.js
+++ b/server/models/subscription.js
@@ -1,0 +1,19 @@
+module.exports = (sequelize, DataTypes) => {
+  const Subscription = sequelize.define('Subscription', {
+    type: DataTypes.ENUM('Subscription'),
+    field: DataTypes.STRING,
+    topic: DataTypes.STRING,
+    filter: DataTypes.JSON
+  })
+
+  Subscription.associate = (models) => {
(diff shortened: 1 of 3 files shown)


server/models/subscription.js
```diff
@@ -0,0 +1,19 @@
+module.exports = (sequelize, DataTypes) => {
+    const Subscription = sequelize.define("Subscription", {
+        type: DataTypes.ENUM("Subscription"),
+        field: DataTypes.STRING,
+        topic: DataTypes.STRING,
+        filter: DataTypes.JSON
+    });
+
+    Subscription.associate = models => {
```
pb82: We do the associations slightly different, look here: https://github.com/aerogear/data-sync-ui/blob/master/server/models/index.js#L12

psturc: oops, sorry I'll fix that asap

psturc:
Verified this by provisioning data-sync-apb with new data-sync-ui image to internal remote cluster, which means https://github.com/aerogearcatalog/data-sync-apb/pull/10 is fixed.

However code review is still needed.

PR #77, opened 2018-07-31
Resolvers tab

pb82:
## Motivation

Implements the left side of the resolvers tab (structure view & resolver selection)

## Checklist:

## Progress

- [x] Finished task

## Additional Notes

When i click on the resolver of a query or mutation that has arguments, the arguments will expand. That should be improved but can be done in another PR.


ui/components/resolvers/ResolversList.js
```diff
@@ -0,0 +1,159 @@
...
+                            kind={kind}
+                            item={field}
+                            resolvers={data}
+                            onClick={onClick}
+                        />
+                    );
+                });
+
+                return (
+                    <div className={style["structure-content"]}>
```
cfoskin: Just wondering about the change between this and `styles.something`, which should we use?

pb82: Use the dot notation when possible. In this case it's not possible because JS properties can't have a `-` in their name.

cfoskin:
@pb82  verified that behaviour is as expected, Code all looks ok but pr is a bit big for me to give code review on. I will let @josemigallas do that.

ui/components/resolvers/ResolversList.js
```diff
@@ -0,0 +1,170 @@
...
+                            { list }
+                        </ListView>
+                    </div>
+                );
+            }}
+        </Query>
+    );
+};
+
+const renderCustomType = (schemaId, item, text, kind, onClick) => {
```
josemigallas: I'd extract all this render methods into their own components `CustomTypeItem`, `QueryItem`, etc

ui/components/resolvers/ResolversListItem.js
```diff
@@ -0,0 +1,152 @@
...
+            description={<span>{fieldsText}</span>}
+            hideCloseIcon
+        >
+            { renderFields(type, fields, resolvers, onClick) }
+        </ListViewItem>
+    );
+};
+
+const ResolversListItem = props => {
+    switch (props.kind) {
```
josemigallas: We would avoid so much hassle if we define a component for each kind of item (see comment above).

ui/helper/GraphQLWellKnownTypes.js
```diff
@@ -0,0 +1,14 @@
+export const wellKnownTypes = [
```
josemigallas: I wonder if it would make more sense to define these in `ui/graphql/types/WellKnownTypes.js` instead.

ui/components/resolvers/ResolversListItem.js
pb82: In the above case i agree. In this case not sure, the differences are not that big and i would rather not have two components one for the above comment (CustomTypeItem) and one here (CustomTypeListItem)

PR #78, opened 2018-08-01
[WIP] Refactor tests to jest

josemigallas:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7725

## What
Adds a boilerplate for UI testing using Jest and Enzyme. Also refactors server tests and removes mocha in favor of Jest as a single framework.

## Why
UI has to have a basic test coverage.

## How
Test framework is Jest. Enzyme is used to render components and work with them.

## Verification Steps
`$ npm test`


psturc:
👀

server/server.js
```diff
@@ -20,52 +16,25 @@ App.use(json());
...
 
-exports.stop = () => {
+exports.stop = callback => {
     log.info("Shutting down UI server");
-    server.close();
     stopNotifications();
-    database.close();
+    database.close()
+        .then(() => {
+            server.close(() => callback());
```
psturc: This throws and error: https://gist.github.com/psturc/76381f9156cfa899f977614fdf0f338d (after starting and stopping the server)

test/server/server.test.js
```diff
@@ -0,0 +1,233 @@
...
+            await graphql(Schema, CREATE_DATA_SOURCE, root, null, dataSource);
+
+            await graphql(Schema, UPSERT_RESOLVER, root, null, {
+                schemaId: 1, dataSourceId: 1, ...resolver
+            });
+
+            const fetch = await graphql(Schema, GET_RESOLVERS, root, null, {
+                schemaId: 1, type: resolver.type
+            });
+            expect(fetch.data.resolvers).toHaveLength(1);
```
psturc: Isn't this the same test as the one above except for the final fetch and expectation? Could it be merged into one test?

test/server/server.test.js
```diff
@@ -0,0 +1,233 @@
...
+
+            const fetch = await graphql(Schema, GET_DATA_SOURCES, root);
+            expect(fetch.data.dataSources).toHaveLength(1);
+            expect(fetch.data.dataSources[0]).toEqual({ id: 1, ...dataSource });
+        });
+
+        it("should delete an existing data source", async () => {
+            await graphql(Schema, CREATE_DATA_SOURCE, root, null, dataSource);
+
+            let fetch = await graphql(Schema, GET_DATA_SOURCES, root);
```
psturc: I'm wondering if we could make use of (and that means also test) returned values from the mutations instead of repeating `GET_*` queries in tests?
i.e.
```
const fetch = await graphql(Schema, CREATE_DATA_SOURCE, root, null, dataSource);
expect(fetch.data.createDataSource).toEqual({ id: 1, ...dataSource });
```

test/server/server.test.js
josemigallas: Good point.

test/server/server.test.js
josemigallas: That is correct and it's actually how I was doing it before. However I think it's more accurate to tests against the DB final state. I might happen that the mutation returned some value that hasn't been added in the DB as we expected. In any case I would check *both*, are those mutation results normally used?

psturc: I think it generally makes sense to use returned values from performed mutation it in the app. Just my assumption (correct me if I'm wrong), but I see two benefits:
1. Amount or requests made by client app is minimised
2. Content is fetched faster (DB response time could be slow)

Anyway, testing both scenarios seems like a good idea to me.

pb82: @psturc i'm sure there are situations where using the return value of a mutation instead of an extra request makes sense. But we don't really do that very often and Apollo seems to encourage a different way. For example after adding a new data source, instead of adding the result of the mutation manually and re-rendering, we just use Apollos `refetchQuey` option. Performance wise it's fine: it doesnt fetch anything from the backend because the result of the last query is still in the cache and the mutation also updated the cache with the new data source.

And from a testing perspective we don't really care how many requests are made. Although i agree that it would make sense to at least check that the return values of the mutations are the ones we expect.

psturc: @pb82 ok thanks for explanation. So since we're not using mutation results anyhow in our app, it's probably not necessary to test them

PR #79, opened 2018-08-01
[WIP] AEROGEAR-7631 - validate data source and show the result in the UI

aliok:
JIRA: https://issues.jboss.org/AEROGEAR-7631

Still a WIP, I have a few questions.

Overall idea:
* add a `getDataSourceTestResult` query in the schema
* resolver for that tests the connection with the provided config and it returns `{status:true}` or `{status: false, message: "Cannot connect: CONNREFUSED or SOMETHINGELSE"}`
* A new test button in the data source dialog, only shown in Postgres
* Test button sends the config in the form to that `getDataSourceTestResult` query. It doesn't send an id of existing data source because users would want to test connection before saving the data source
* In the dialog, show either "Connected successfully" or "Cannot connect: CONNREFUSED or SOMETHINGELSE" message at the top

I am only having trouble with how to call that query. Please see my inline GH comment.

cc @josemigallas @cfoskin @pb82

### DIFF
--- a/ui/components/data-sources/BaseDataSourceDialog.js
+++ b/ui/components/data-sources/BaseDataSourceDialog.js
@@ -17,6 +17,8 @@ import some from "lodash.some";
 import { InMemoryOptions, PostgresOptions } from "./options";
 import { DataSourceType } from "../../graphql/types/DataSourceType";
 import { Validators, Validate } from "../../helper/Validators";
+import TestDataSource from "../../graphql/TestDataSource.graphql";
+import ApolloClient from "apollo-boost"
 
 class BaseDataSourceDialog extends Component {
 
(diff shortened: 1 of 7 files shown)


ui/components/data-sources/BaseDataSourceDialog.js
```diff
@@ -151,9 +151,33 @@ class BaseDataSourceDialog extends Component {
...
+            .catch((message) => this.setState({ err: message }));
+    }
+
+    testDataSource() {
+        const { name, type } = this.state;
+
+        const config = { options: this.getConfigByType(type) };
+        // TODO: what here?
+        // TODO: how to call `getDataSourceTestResult` query?
+
```
aliok: I don't know how to call the `getDataSourceTestResult` query within this component.

@josemigallas any help?

Normally, data source dialog subclass instances are bound to a single mutation. But, in this case, this is an additional functionality which has nothing to do with the mutation.

I tried this:
```
import gql from "graphql-tag";

client
  .query({
    query: gql`
      {
        rates(currency: "USD") {
          currency
        }
      }
    `
  })
  .then(result => console.log(result));
```
which is simply using the ApolloClient with nothing specific to React, but I don't know how to get a reference to `client`. 
@josemigallas does this look good to you? or, is there any other approach you would recommend?

pb82: @aliok One possibility is to use a `<Query>` tag. See here for an example: https://github.com/aerogear/data-sync-ui/blob/master/ui/components/resolvers/ResolversList.js#L94

aliok: wow that looks ugly :smiling_imp:  :laughing: 
will give it a try

pb82: Another possibility is to import `graphql` from `react-apollo` and use this. I'll give it a try and let you know.

josemigallas:
@aliok Regardless of the validity of code at this point, this is the perfect example for creating a new component.

ui/components/data-sources/BaseDataSourceDialog.js
```diff
@@ -235,6 +267,15 @@ class BaseDataSourceDialog extends Component {
                 </Modal.Body>
 
                 <Modal.Footer>
+                    {type === DataSourceType.Postgres && (
+                        <Button
+                            bsStyle="info"
+                            onClick={() => this.onTest()}
+                            disabled={submitButtonDisabled}
```
aliok: I will keep this even though user doesn't have to actually type data source name to test connection.
If fields are empty, test function doesn't work properly (uses default like localhost:5432, user:aliok password:none). So, let's make sure UI asks all fields to be filled, like "submit" button. 

A better way would be to use the same validation rules, but not require name input to be filled. I don't think this is worth the effort.

pb82: I'm happy with this

PR #84, opened 2018-08-03
Kill connection properly during Postgres connection test

aliok:
A small improvement on killing the connection after Postgres connection is tested.
Based on the discussion here: https://github.com/brianc/node-postgres/issues/397#issuecomment-409894050


aliok:
cc @pb82 no JIRA for this one

aliok:
@pb82 linter is now happy

PR #88, opened 2018-08-06
Audit logs

aliok:
JIRA: https://issues.jboss.org/browse/AEROGEAR-7639

What:
* Added a special logger function that adds a tag "AUDIT" to the output JSON.
* That tag will be used to search things when we centralize logging on OpenShift with EFK.
* Used that function in create/update/delete operations for resources (schema, data source, resolver)
* No user/client information is logged ATM as we don't have auth yet.

How to verify:
* Just start the server, execute bunch of operations and check out the STDOUT. There will be some logs with tag AUDIT.

Note:
I used my best judgement on what to log (e.g. logging config for data sources) but I am open to discussions.

### DIFF
--- a/server/gql/schema.js
+++ b/server/gql/schema.js
@@ -1,5 +1,5 @@
 const { buildSchema } = require("graphql");
-const { log } = require("../logger");
+const { log, auditLog } = require("../logger");
 const { dataSource, database, supportsiLike, schema, resolver } = require("../models");
 const { compileSchemaString, formatGraphqlErrors } = require("./helper");
 const { publish, DEFAULT_CHANNEL } = require("../configNotifiers/configNotifierCreator");
@@ -77,6 +77,12 @@ const createDataSource = async ({ name, type, config }) => {
     });
 
     publish(DEFAULT_CHANNEL, { reload: "DataSource" });
+    auditLog({
+        operation: "createDataSource",
+        name,
+        type,
+        config
+    });
     return created;
 };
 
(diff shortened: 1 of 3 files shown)


server/gql/schema.js
```diff
@@ -117,11 +123,24 @@ const upsertResolver = async ({
...
     if (id) {
         const updated = await resolver.findById(id);
-        return updated.update(properties);
+        result = updated.update(properties);
+    } else {
+        result = resolver.create(properties);
     }
-    return resolver.create(properties);
+
+    publish(DEFAULT_CHANNEL, { reload: "Resolver" });
```
aliok: @pb82 as I told you on friday, publishing on upserting a resolver was missing. I added it here.

server/logger.js
```diff
@@ -1,7 +1,18 @@
 const log = require("pino")();
 const expressPino = require("express-pino-logger")({ level: "silent", logger: log });
```
pb82: Not a change in your PR but i wonder why we configure expressPino when we just set it to `silent` with no option to turn it on.

aliok: :) no idea LOL

server/logger.js
```diff
@@ -1,7 +1,18 @@
...
 
+const auditLogger = log.child({ tag: "AUDIT" });
+
+const auditLogEnabled = process.env.AUDIT_LOGGING !== "false" && process.env.AUDIT_LOGGING !== false;
```
pb82: I would move this to https://github.com/aerogear/data-sync-ui/blob/master/server/config.js and import it from there so that we have all env vars in one file.

aliok: Addressed this here: https://github.com/aerogear/data-sync-ui/pull/89

PR #89, opened 2018-08-06
Audit logs - move related config to config.js

aliok:
Move audit log related config to config.js


aliok:
Thanks Peter. Forgot about this PR :)

PR #80, opened 2018-08-01
Implement Resolver Detail

josemigallas:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7724

## Description
This is a work in progress, do not merge and do not review exhaustively.

PR #90, opened 2018-08-06
Resolver Detail Hooks

josemigallas:
~~NOTE: built on top of #80, do not merge until #80 it is merged.~~

This PR adds:

- preHook and postHook validation
- dummy preHook and postHook URL verification (waiting for https://github.com/aerogear/data-sync-server/pull/46)

TODO:

- [x] Disable hooks when resolver is for a Query
- [x] Implement hook test button (depends on waiting for https://github.com/aerogear/data-sync-server/pull/46)


ui/components/resolvers/ResolverDetail.js
```diff
@@ -55,6 +62,44 @@ class ResolverDetail extends Component {
...
+        const newValidations = { ...validations, dataSource: dsValidation };
+
+        this.setState({ validations: newValidations });
+        this.updateResolver({ DataSource });
+    }
+
+    onPreHookChange(preHook) {
+        const preHookValidation = Validate([
+            Validators.String.minLength(1), preHook,
+            Validators.String.maxLength(255), preHook,
```
pb82: Could be too small for a URL. The limit seems to be closer to 2000 chars: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

psturc:
This is more related to #80, but should we warn user about unsaved changes in **Resolvers** tab also when the page is being reloaded?

ui/helper/Validators.js
```diff
@@ -15,6 +15,15 @@ const Validators = {
     },
     Boolean: {
         valid: b => typeof b === typeof true
+    },
+    URL: {
+        valid: u => {
+            try {
+                return !!new URL(u);
```
camilamacedo86: Is it correct `! !`?

josemigallas: That's correct, it's a quick way to parse any value into a `boolean`.

ui/graphql/UpsertResolver.graphql
```diff
@@ -4,15 +4,19 @@ mutation UpsertResolver(
...
-    $responseMapping: String!
+    $responseMapping: String
 ) {
     upsertResolver(
         id: $id,
         schemaId: $schemaId,
         dataSourceId: $dataSourceId,
         type: $type,
         field: $field,
+        preHook: $preHook,
```
pb82: You need to add `preHook` and `postHook` to the result of the mutation too. Otherwise they will be emptied after save.

PR #91, opened 2018-08-06
Fix validators

josemigallas:
Adds tests for `Validator.js` and also fixes some validations.

### DIFF
--- a/test/ui/helper/Validators.test.js
+++ b/test/ui/helper/Validators.test.js
@@ -142,3 +142,15 @@ describe("URL", () => {
         expect(Validate([Validators.URL.valid, {}])).toEqual("error");
     });
 });
+
+describe("Password", () => {
+    it("should validate a password as long as it is a string", () => {
+        expect(Validate([Validators.Password.valid, ""])).toEqual("success");
+        expect(Validate([Validators.Password.valid, "123123"])).toEqual("success");
+        expect(Validate([Validators.Password.valid, "null"])).toEqual("success");
+        expect(Validate([Validators.Password.valid, null])).toEqual("error");
+        expect(Validate([Validators.Password.valid, undefined])).toEqual("error");
+        expect(Validate([Validators.Password.valid, NaN])).toEqual("error");
+        expect(Validate([Validators.Password.valid, 1234])).toEqual("error");
+    });
+});
--- a/test/ui/helper/Validators.test.js
+++ b/test/ui/helper/Validators.test.js
@@ -82,15 +82,18 @@ describe("String", () => {
 
 describe("Number", () => {
     it("should validate a natural number", () => {
+        expect(Validate([Validators.Number.natural, "4"])).toEqual("success");
+        expect(Validate([Validators.Number.natural, "0"])).toEqual("success");
+
         expect(Validate([Validators.Number.natural, 4])).toEqual("error");
         expect(Validate([Validators.Number.natural, 0])).toEqual("error");
         expect(Validate([Validators.Number.natural, -0])).toEqual("error");
+        expect(Validate([Validators.Number.natural, NaN])).toEqual("error");
+        expect(Validate([Validators.Number.natural, undefined])).toEqual("error");
+        expect(Validate([Validators.Number.natural, null])).toEqual("error");
         expect(Validate([Validators.Number.natural, 1.1])).toEqual("error");
         expect(Validate([Validators.Number.natural, -1])).toEqual("error");
         expect(Validate([Validators.Number.natural, -1.1])).toEqual("error");
-
-        expect(Validate([Validators.Number.natural, "4"])).toEqual("success");
-        expect(Validate([Validators.Number.natural, "0"])).toEqual("success");
         expect(Validate([Validators.Number.natural, "-0"])).toEqual("error");
         expect(Validate([Validators.Number.natural, "1.1"])).toEqual("error");
         expect(Validate([Validators.Number.natural, "-1"])).toEqual("error");
--- a/test/ui/helper/Validators.test.js
+++ b/test/ui/helper/Validators.test.js
@@ -3,9 +3,9 @@ import { Validate, Validators } from "../../../ui/helper/Validators";
 it("should correctly validate a bunch of cases", () => {
     const result = Validate([
         Validators.String.nonBlank, "some random string",
-        Validators.Number.natural, 123,
+        Validators.Number.natural, "123",
         Validators.URL.valid, "http://localhost",
-        Validators.Port.valid, 123
+        Validators.Port.valid, "123"
     ]);
 
     expect(result).toEqual("success");
@@ -16,7 +16,7 @@ it("should find a wrong value from a bunch of validations", () => {
         Validators.String.nonBlank, "some random string",
         Validators.Number.natural, null,
         Validators.URL.valid, "http://localhost",
-        Validators.Port.valid, 123
+        Validators.Port.valid, "123"
     ]);
 
     expect(result).toEqual("error");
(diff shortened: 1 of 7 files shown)


ui/helper/Validators.js
```diff
@@ -3,12 +3,12 @@ const Validators = {
...
         nonBlank: s => s
                 && typeof s === typeof ""
-                && s.length > 0,
+                && s.trim().length > 0,
         minLength: l => s => Validators.String.nonBlank(s) && s.length >= l,
         maxLength: l => s => Validators.String.nonBlank(s) && s.length < l
     },
     Number: {
-        natural: n => !Number.isNaN(n) && parseInt(n, 10) > 0
+        natural: n => Number.isInteger(n) && n >= 0
```
pb82: Hmm, not sure why i went with `isNaN` but your solution seems fine

josemigallas: Because you didn't have tests to prove otherwise ;)

josemigallas:
@pb82 It fails because it's passing a String, not an actual number. Do we want to include strings as valid numbers?
https://github.com/aerogear/data-sync-ui/pull/91/files#diff-fbcf5225cb21f2fbdb020c5cc734f1ddR52

josemigallas:
@pb82 something's wrong with the new validation "details" mode -> https://github.com/aerogear/data-sync-ui/blob/a0eb1f37ff8422d21ac428abedea75e04d50f211/test/ui/helper/Validators.test.js#L28-L38
And port is always "success" in the Postgres version of `AddDataSourceDialog`.

test/ui/helper/Validators.test.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(Validate([Validators.String.nonBlank, ""])).toEqual("error");
+        expect(Validate([Validators.String.nonBlank, "   "])).toEqual("error");
+    });
+});
+
+describe("Number", () => {
+    it("should validate a natural number", () => {
+        expect(Validate([Validators.Number.natural, "4"])).toEqual("success");
+        expect(Validate([Validators.Number.natural, "0"])).toEqual("success");
+        // Anything with a minus (even zero) is not considered a natural number
```
josemigallas: Test case explains itself, no need for comments.

test/ui/helper/Validators.test.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(Validate([Validators.String.nonBlank, undefined])).toEqual("error");
+        expect(Validate([Validators.String.nonBlank, NaN])).toEqual("error");
+        expect(Validate([Validators.String.nonBlank, ""])).toEqual("error");
+        expect(Validate([Validators.String.nonBlank, "   "])).toEqual("error");
+    });
+});
+
+describe("Number", () => {
+    it("should validate a natural number", () => {
+        expect(Validate([Validators.Number.natural, "4"])).toEqual("success");
```
josemigallas: Don't remove valid cases such as actual numbers `4`, `-1`, etc. Plus, you're validating a `String`, not a `Number` here, it's a little bit misleading.

test/ui/helper/Validators.test.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(Validate([Validators.Number.natural, "-0"])).toEqual("error");
+        expect(Validate([Validators.Number.natural, "1.1"])).toEqual("error");
+        expect(Validate([Validators.Number.natural, "-1"])).toEqual("error");
+        expect(Validate([Validators.Number.natural, "-1.1"])).toEqual("error");
+    });
+});
+
+describe("Port", () => {
+    it("should validate a valid port", () => {
+        expect(Validate([Validators.Port.valid, "0"])).toEqual("success");
```
josemigallas: Again, don't remove the valid cases I added. Is `123` a valid port? If Port has to be a String, then add other cases like `"aaa"`.

ui/components/data-sources/AddDataSourceDialog.js
```diff
@@ -22,7 +22,13 @@ const INITIAL_STATE = {
     validations: {
         name: null,
         type: "warning",
-        options: "success"
+        options: "error",
+
+        // Stores validation states for postgres
+        // properties
+        postgresDetails: {
+            port: "success"
```
josemigallas: Why is this one separated? What about the other forms?

ui/components/data-sources/BaseDataSourceDialog.js
```diff
@@ -67,22 +67,17 @@ class BaseDataSourceDialog extends Component {
...
-        Validate([
+        const postgresDetails = {};
+
+        const options = Validate([
             Validators.URL.valid, url, "url",
             Validators.String.nonBlank, database, "database",
             Validators.String.nonBlank, username, "username",
-            Validators.Port.valid, port, "port"
-        ], validationDetails);
+            Validators.Number.natural, port, "port"
```
josemigallas: Why do we have `Validators.Port` then?

ui/components/data-sources/BaseDataSourceDialog.js
```diff
@@ -67,22 +67,17 @@ class BaseDataSourceDialog extends Component {
...
+        ], postgresDetails);
 
         const { validations } = this.state;
-        const newValidations = { ...validations, ...validationDetails };
+        const newValidations = { ...validations, postgresDetails, options };
```
josemigallas: What is `options` and why don't we include those under `postgresDetails`?

ui/components/data-sources/BaseDataSourceDialog.js
```diff
@@ -200,7 +195,17 @@ class BaseDataSourceDialog extends Component {
     render() {
         const { visible } = this.props;
         const { name, type, err, success, validations } = this.state;
-        const submitButtonDisabled = some(validations, s => !s || s === "error");
+
+        // InMemory options are always correct: there is only one options which
```
josemigallas: This validation logic belongs to `onTypeChange`.

ui/components/data-sources/options/PostgresOptions.js
```diff
@@ -63,7 +63,7 @@ const PostgresOptions = ({ options, onOptionsChange, isDisabled, validations })
                 />
             </Col>
         </FormGroup>
-        <FormGroup controlId="password" validationState={validations.password}>
+        <FormGroup controlId="password" validationState="success">
```
josemigallas: Don't restrain extensibility, we don't know what's the future validation for password. Let's better delegate this value to onPasswordChange or in any case, the initial state.

ui/components/data-sources/AddDataSourceDialog.js
pb82: We only have a default value for the port and want it to be valid without running the validators first.

ui/components/data-sources/BaseDataSourceDialog.js
pb82: `options` is the overall validation result that determines if the `Ok` button will be disabled. `postgresOptions` contains the validation results for the postgres options and is only used to display the validation state in the postgres options component. We can't include them in `validations` directly because based on the selected type they are available or not.

ui/components/data-sources/BaseDataSourceDialog.js
pb82: Won't work. We want the submit button for `InMemory` to be enabled by default without changing anything first (because there are no options to check: all options are always correct for in mem).

ui/components/data-sources/AddDataSourceDialog.js
pb82: Other forms (in mem) don't have default options (or validation state) yet.

ui/components/data-sources/options/PostgresOptions.js
```diff
@@ -1,83 +1,141 @@
...
+};
+
+class PostgresOptions extends Component {
+
+    constructor(props) {
+        super(props);
+        this.state = INITIAL_STATE;
+    }
+
+    onUrlChange(url) {
```
pb82: Why not one `onChange` function that handles all validations? Bit less performant but fewer lines of code.

PR #92, opened 2018-08-07
show validations per field for postgres

pb82:
## Motivation

Show the validation state for each field when the postgres options are selected. This is achieved by extending the `Validate` function and allowing it to pass in field names and validation details.


cfoskin:
@pb82  is anything basically valid once we have a datasource name and the port is correct format?

PR #94, opened 2018-08-08
added delete dialog

cfoskin:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7594

## What
Add delete functionality for the resolver

## Why
The user should be able to delete resolvers

## How
Add a message dialog to pop up when they click the new delete button which will prompt them to delete the resolver and show the id of it also

## Verification Steps
Create a datasource
Create a schema
Create a resolver
Click that resolver and choose to delete it
It should delete it and show no resolver on the query or mutation that you attached the resolver to

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/ui/components/resolvers/DeleteResolverDialog.js
+++ b/ui/components/resolvers/DeleteResolverDialog.js
@@ -26,7 +26,7 @@ const DeleteResolverDialog = ({
         });
     };
 
-    const title = `Delete Resolver ${resolver.id}`;
+    const title = "Delete Resolver";
     const primaryContent = (
         <p style={{ fontSize: 18 }}>
             Are you sure you want to delete this resolver?
--- a/ui/components/resolvers/DeleteResolverDialog.js
+++ b/ui/components/resolvers/DeleteResolverDialog.js
@@ -1,0 +1,55 @@
+import React from "react";
+
+import { MessageDialog } from "patternfly-react";
+import { graphql } from "react-apollo";
+import DeleteResolver from "../../graphql/DeleteResolver.graphql";
+import GetResolvers from "../../graphql/GetResolvers.graphql";
+
+const DeleteResolverDialog = ({
+    showModal,
+    resolver,
+    mutate,
+    onDelete,
+    onClose
+}) => {
+    const removeResolver = () => {
+        const resolverId = resolver.id;
+        const { schemaId, type } = resolver;
+        mutate({
+            variables: { resolverId },
+            refetchQueries: [{ query: GetResolvers, variables: { schemaId, type } }]
+        }).then(() => {
+            onDelete();
+        }).catch(err => {
+            console.error("Error", err);
+        });
+    };
+
+    const title = `Delete Resolver ${resolver.id}`;
+    const primaryContent = (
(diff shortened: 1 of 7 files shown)


ui/components/resolvers/ResolverDetail.js
```diff
@@ -182,9 +192,20 @@ class ResolverDetail extends Component {
...
+                    >
+                        Delete Resolver
+                    </Button>
                 </div>
 
+                <DeleteResolverDialog
+                    showModal={showDeleteModal}
+                    resolver={resolver}
+                    onClose={() => this.setState({ showDeleteModal: false })}
+                    onDelete={() => this.onDeleteResolver()}
```
pb82: Do we really need two different callbacks?

ui/components/resolvers/ResolverDetail.js
```diff
@@ -182,9 +192,20 @@ class ResolverDetail extends Component {
                     >
                         Save
                     </Button>
-                    {/* <Button bsStyle="default" onClick={() => console.log("cancel")}>Cancel</Button> */}
+                    <Button
+                        bsStyle="default"
```
pb82: There is a button type `danger` i think. It would be more suitable here.

ui/components/resolvers/ResolverDetail.js
```diff
@@ -182,9 +192,20 @@ class ResolverDetail extends Component {
                     >
                         Save
                     </Button>
-                    {/* <Button bsStyle="default" onClick={() => console.log("cancel")}>Cancel</Button> */}
+                    <Button
```
pb82: This button needs to be disabled when delete is not possible (i.e. before saving a new resolver)

ui/components/resolvers/DeleteResolverDialog.js
```diff
@@ -0,0 +1,53 @@
...
+            variables: { resolverId },
+            refetchQueries: [{ query: GetResolvers, variables: { schemaId, type } }]
+        }).then(() => {
+            onDelete();
+        }).catch(err => {
+            console.error("Error", err);
+        });
+    };
+
+    const title = `Delete Resolver ${resolver.id}`;
```
pb82: I would remove the Id from the title

ui/components/resolvers/ResolverDetail.js
cfoskin: The onclose one only hides the modal. The ondelete happens after the delete resets the entire state which basically unselects the resolver and shows the empty screen on the right. I think this is actually ok no what do you think? if it is I will just use the one function and reset the state If they press cancel (which also closes the modal)

pb82: It's ok, two callbacks make sense here.

cfoskin: 👍

cfoskin:
@pb82 I can verify that your additional changes all work as expected, Do we want to merge this now?

pb82:
@cfoskin would be good to add a little ui test for this. Making sure that the button is disabled / visible, etc.

PR #96, opened 2018-08-09
dont force close the connection

pb82:
## Motivation

`client.connection.stream.destroy();` was throwing un unhandled exception. Just using `client.end()` seems to work.

### DIFF
--- a/server/dataSourceValidator/validators/postgres.js
+++ b/server/dataSourceValidator/validators/postgres.js
@@ -63,18 +63,7 @@ module.exports = async config => {
         return { status: true };
     } finally {
         if (client) {
-            // force kill the connection if it is open or end it if it is not
-            // see https://github.com/brianc/node-pg-pool/blob/277dc508daea03a8f6c0bcc3c534cab5b2501b12/index.js#L208-L209
-            try {
-                // do not block on ending the client
-                if (client.connection) {
-                    client.connection.stream.destroy();
-                } else {
-                    client.end();
-                }
-            } catch (ex) {
-                // do nothing even when there's an error.
-            }
+            client.end();
         }
     }
 };


aliok:
@pb82 

I got this code piece from https://github.com/brianc/node-pg-pool/blob/277dc508daea03a8f6c0bcc3c534cab5b2501b12/index.js#L208-L209
and thought it is the best way.

I checked `client.end()` function's impl, and got confused.
https://github.com/brianc/node-postgres/blob/master/lib/client.js#L384

```
Client.prototype.end = function (cb) {
  this._ending = true
  if (this.activeQuery) {
    // if we have an active query we need to force a disconnect
    // on the socket - otherwise a hung query could block end forever
    this.connection.stream.destroy(new Error('Connection terminated by user'))
    return cb ? cb() : Promise.resolve()
  }
  if (cb) {
    this.connection.end()
    this.connection.once('end', cb)
  } else {
    return new global.Promise((resolve, reject) => {
      this.connection.end()
      this.connection.once('end', resolve)
    })
  }
}
```

I am afraid, if we just call `client.end()`, we might end up waiting forever when the connection is currently hung. Similar to comment `a hung query could block end forever`.

I would just be extra cautious and do what node-pg-pool does (dependency of node-postgres)

aliok:
so, more concrete suggestion: wrap destroy call in try/catch

pb82:
@aliok But it already is wrapped in a try/catch. Strange thing is that this still returns a successful result to the UI and only crashes afterwards.

aliok:
@pb82 yeah, I got what you mean. But I don't understand how this can throw an unhandled exception :)

In what case do you see this error?

pb82:
@aliok The server crashes when i test a valid connection (it only crashes after returning the result to the UI though).

aliok:
ok, go and merge this then.

PR #104, opened 2018-08-15
Aerogear 7761

StephenCoady:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7761

## What
Added publish field to schema
Create incoming subscription if required

@darahayes I had the json checks in but removed them because on second thoughts I realised it would mean going against the wording of the ticket (it states a string and I would need to change the data type to JSON). I don't think this is what was required for now. It is easy to add this back if required. Can you take a look if you get a chance.

### DIFF
--- a/server/gql/schema.js
+++ b/server/gql/schema.js
@@ -33,6 +33,7 @@ const Schema = buildSchema(`
             postHook: String
             requestMapping: String!
             responseMapping: String
+            publish: String
         ): Resolver
         deleteResolver(id: Int!): Resolver
     },  
@@ -61,6 +62,7 @@ const Schema = buildSchema(`
         requestMapping: String!
         responseMapping: String!
         DataSource: DataSource!
+        publish: String
     },
     type DataSourceTestResult {
         status: Boolean!
@@ -115,7 +117,7 @@ const getOneResolver = ({ id }) => resolver.findById(id, {
 
 const upsertResolver = async ({
     id, schemaId, dataSourceId, type, field, preHook = "", postHook = "",
-    requestMapping, responseMapping = ""
+    requestMapping, responseMapping = "", publish
 }) => {
     const properties = {
         GraphQLSchemaId: schemaId,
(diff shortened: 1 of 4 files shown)


server/gql/schema.js
```diff
@@ -124,13 +126,24 @@ const upsertResolver = async ({
...
+        field,
+        publish: publisher
     };
 
     let result;
     if (id) {
         const updated = await resolver.findById(id);
         result = updated.update(properties);
+        if (properties.publish) {
+            const subExists = await subscription.findOne({ where: { field: properties.publish } });
```
darahayes: So this is actually a pretty cool candidate for the `findOrCreate`. You can see this in the sequelize docs here: http://docs.sequelizejs.com/manual/tutorial/models-usage.html

Not necessary though

pb82: Looks good, but maybe extract it into a new function like `checkAndUpdateSubscriptions`

camilamacedo86: In 2 functions check and update.  It is not a good practice function as credateDeleteUpdate. Each one should be responsible for one action.

darahayes: @pb82 @camilamacedo86 @StephenCoady you wouldn't need to worry about this at all if you just use `model.findOrCreate` because it would be a single line of code.

pb82: Hmm, it's kind of two unrelated things: 1) updating the resolver and 2) possibly creating a subscription. But if we can reduce it to one line using `findOrCreate` then it should be fine to keep it in `upsertResolver`.

server/gql/schema.js
```diff
@@ -33,6 +33,7 @@ const Schema = buildSchema(`
             postHook: String
             requestMapping: String!
             responseMapping: String
+            publisher: String
```
josemigallas: I think it has to be `publish` here as well.

josemigallas:
@StephenCoady Would you mind adding a couple of tests for this feature?

StephenCoady:
@josemigallas mind taking a look now?

PR #109, opened 2018-08-17
Review styles to match given mockups.

josemigallas:
## Motivation
The UI should look as much as possible like the mockups given by design/UX responsibles.

## What

- Reviews CSS and change what's necessary.
- Moves styles from main.css into specific .css files per component, when possible.
- Refactors how classes are assigned into a more JS friendly way (from `styles["class"]` to styles.class)
- Refactors in resolvers list components, for simplicity and easy of readability.

## Verification Steps

- Verify UI resembles [mockups](https://redhat.invisionapp.com/share/SVMB0QW6MR9#/screens) as much as possible. 
- Due to refactorizations, verify Resolvers tab keeps working as expected.


psturc:
👀

josemigallas:
@psturc definitely, I will create a ticket now so that it gets tracked.

josemigallas:
@cfoskin 
> should we re-label the button to download compiled schema to Export ?
I'm fine with Export but maybe it's not clear enough what is being exported exactly, I think @pb82 changed it for that very reason.

> Do we also want to add the info for resolvers added to queries and mutations? like 0/2 or 1/2 etc...
Definitely, I will create a ticket to track it, I think it would be too much work already on the same PR.

> We should also make the save button blue on the resolvers tab (to be the same as mock ups)
It is blue but only when it's enabled, just like the on in Schema tab's toolbar.

PR #110, opened 2018-08-17
Subscriptions dropdown

josemigallas:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7760

## What
Adds a dropdown to add a subscription to a Resolver.

## Why
Because the JIRA says so.

## How
Subscriptions are fetched inside the component, right now the query does not work. (probably need some changes to DB related logic)

## Verification Steps

1. Go to Resolvers and select a resolver.
2. Verify there is a dropdown for subscriptions.
    - If there is no subscriptions, a message is shown instead of a dropdown.
    - If there are subscriptions but none is selected, show form empty.

## Checklist:

## Progress

- [x] Finished UI
- [x] Fix DB related logic (getSubscriptions not working)
- [x] Add tests for the UI
- [x] Add some test for getSubscriptions


ui/components/resolvers/SubscriptionsDropDown.js
```diff
@@ -0,0 +1,62 @@
...
+        <Col sm={2} className={controlLabel}>Subscription</Col>
+        <Col sm={6}>
+            <Query query={GetSubscriptions} variables={undefined}>
+                {({ loading, error, data }) => {
+                    if (!data || !data.subscriptions || !data.subscriptions.length) {
+                        return <FormControl.Static>Please first create a subscription</FormControl.Static>;
+                    }
+
+                    const options = [
+                        <MenuItem key="empty_subscription" eventKey={undefined} />,
```
jhellar: It is hard to select the empty option from dropdown. Maybe add a text to it (e.g. "None")?

psturc:
👀

psturc:
Looks fine to me. Will do final verification once the issue with `getSubscriptions` is resolved.

wtrocki:
> There are no subscriptions, or a way to create them anyway.

What you mean by this? We have example subscription in memeolist.
This is critical issue for this epic so maybe we could sync for quick demo. WDYT? :>

wtrocki:
The way we read that at the moment do not map the data back from the database. 
So creating subscription gets some strange json:
screen shot 2018-08-27 at 12 43 11 pm

psturc:
👀

psturc:
Two issues seen, both probably related to a single problem:
1. [saving resolver without subscription](https://drive.google.com/file/d/1-Mm0TO0XfJBu-2EJlFoSdT_WLaWdHCun)
2. [removing subscription from resolver](https://drive.google.com/open?id=1k-WQgPMuULQeJfL5anOZjaib0hrTi1Ui)

wtrocki:
@psturc Awesome work! I did not catched that when testing. Like that you went and actually made videos with the problem.

josemigallas:
> The way we read that at the moment do not map the data back from the database. 
So creating subscription gets some strange json:
screen shot 2018-08-27 at 12 43 11 pm

@wtrocki This is as expected. Currently the UI has to consider `publish` as a string that matches Subscriptions' topic property.

wtrocki:
@josemigallas Great to see!

PR #111, opened 2018-08-17
Fixed cursor positioning after hitting tab button

psturc:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7789
## What
Previously, in the schema editor, when you hit TAB button, new tab character was added as expected, but a cursor jumped to the last character of the whole schema
## How
~~By passing a callback to `setState` function, which may not be the best approach.
Unfortunately the only one that worked for me (feel free to suggest a better way)~~
Record position of cursor after Tab key was hit and fix its position in `componentDidUpdate` lifecycle method.
## Verification Steps
Go to schema editor, type couple of lines of text, move cursor to the middle (between characters) and press tab. It should behave like you're used to in common text editor.

### DIFF
--- a/ui/components/common/CodeEditor.js
+++ b/ui/components/common/CodeEditor.js
@@ -22,7 +22,16 @@ const CodeEditor = class extends Component {
             // When the user presses tab we prevent the default propagation, so we need to udpate
             // the state manually here
             this.props.onChange(`${value.substring(0, selectionStart)}\t${value.substring(selectionEnd)}`);
+            this.updateCursorPosition(selectionStart + 1);
         }
+    }
+
+    updateCursorPosition(position) {
+        const editor = this.editor.current;
+        this.setState({}, () => {
+            editor.selectionStart = position;
+            editor.selectionEnd = position;
+        });
     }
 
     onEditorScroll() {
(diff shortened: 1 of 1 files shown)


ui/components/common/CodeEditor.js
```diff
@@ -22,9 +22,18 @@ const CodeEditor = class extends Component {
             // When the user presses tab we prevent the default propagation, so we need to udpate
             // the state manually here
             this.props.onChange(`${value.substring(0, selectionStart)}\t${value.substring(selectionEnd)}`);
+            this.updateCursorPosition(selectionStart + 1);
         }
     }
 
+    updateCursorPosition(position) {
+        const editor = this.editor.current;
+        this.setState({}, () => {
```
josemigallas: What's setState here for?

psturc: My thinking is: if the position of the cursor needs to be updated **after** the component's state is updated ('\t' character is added to textarea), one possible way to do this is via callback passed to `setState` function. This is kinda hacky since we're already updating the textarea with `setState` function triggered by `onChange` event [here](https://github.com/aerogear/data-sync-ui/blob/master/ui/components/schema/SchemaContainer.js#L39)

Honestly I would appreciate some suggestion from React domain experts on how to make this change more elegant and less hacky :P

psturc:
@josemigallas if you're happy with the variable name, let's merge this guy :D

wtrocki:
Schema editor is lacking the basic functionalities - it's not worth to invest time to improve that. We can use existing solutions like prisma

PR #112, opened 2018-08-17
Correct message when all datasources filtered out

wojta:
https://issues.jboss.org/browse/AEROGEAR-7791
## Motivation
Better UX.

## What
Searching the data sources - when searching for an item and it's not found show a `No item found on search` type message.

## Why
As of writing this ticket it shows the blank slate message which could be confusing for the user.

## Verification Steps
1. Display page with datasources, when there is none, it should show `No Data Sources defined` page
2. Add some datasources
3. It should show the datasources in the list
4. Try to filter it by name so the filter no longer matches any of them
5. It should show page `No item found while filtering the data sources` and button `Clear the filter`
6. When pressing `Clear the filter` filter is cleared and all datasources are displayed

## Checklist:

## Progress

- [x] Displaying different messages
- [x] Possibility to clear filter quickly

## Additional Notes

PS.: Add images and/or .gifs to illustrate what was changed if this pull request modify the appearance/output of something presented to the users.


ui/components/common/CommonToolbar.js
```diff
@@ -50,6 +52,8 @@ class CommonToolbar extends Component {
                         type="text"
                         placeholder="Filter by Name"
                         className={toolbarFilter}
+                        value={get(this.props.filter, "name", "")}
```
josemigallas: I think you can avoid installing the new package for this, perhaps destructuring:
```
const { onFilter, filter: { name } } = this.props;
```

ui/components/resolvers/ResolverDetail.js
```diff
@@ -19,6 +19,7 @@ import { DeleteResolverDialog } from "./DeleteResolverDialog";
 import { DefaultEmptyView } from "../common/DefaultEmptyView";
 
 import UpsertResolver from "../../graphql/UpsertResolver.graphql";
+import GetDataSources from "../../graphql/GetDataSources.graphql";
```
josemigallas: I think the branch needs rebasing

ui/components/common/CommonToolbar.js
```diff
@@ -50,6 +52,8 @@ class CommonToolbar extends Component {
...
+                        style={{ height: "26px" }}
```
josemigallas: This should be declared in `.toolbarFilter` css class, but I think it's not needed anyway.

ui/components/data-sources/DataSourcesContainer.js
```diff
@@ -53,6 +53,10 @@ class DataSourcesContainer extends Component {
         this.setState({ showDeleteModal: true, selectedDataSource: dataSource });
     }
 
+    clearFilter() {
+        this.setState({ filter: {} });
```
josemigallas: Why is filter an object?

wojta: It was like that previously, someone probably thought that there might be filters by more criteria.

ui/components/common/CommonToolbar.js
wojta: I understand that. I'm not very experienced with JavaScript, but I think that this is particularly very confusing for someone coming from other languages.

ui/components/data-sources/DataSourcesContainer.js
josemigallas: It's true, my bad!

wtrocki:
@wojta Please rebase and update this. We could merge that after

ui/components/common/CommonToolbar.js
```diff
@@ -47,6 +49,8 @@ class CommonToolbar extends Component {
                 type="text"
                 placeholder="Filter by Name"
                 className={toolbarFilter}
+                value={get(this.props.filter, "name", "")}
```
josemigallas: Why you need `lodash.get`?

ui/components/common/CommonToolbar.js
```diff
@@ -47,6 +49,8 @@ class CommonToolbar extends Component {
...
+                style={{ height: "26px" }}
```
josemigallas: This should removed or at least included in the css class.

ui/components/data-sources/DataSourcesContainer.js
```diff
@@ -34,11 +34,11 @@ class DataSourcesContainer extends Component {
     }
 
     setFilter(nameToFilter) {
-        let name = nameToFilter;
-        if (name === "") {
-            name = undefined;
+        const { name, ...newFilter } = this.state.filter;
```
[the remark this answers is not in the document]
wojta: Actually it's needed for clearing the filter.

ui/components/common/CommonToolbar.js
wojta: ok

ui/components/common/CommonToolbar.js
wojta: Isn't there some other way than decomposing? I really don't like it.

josemigallas: I'm sure you'll get used to it after a few tries! Unfortunately, I don't think installing a whole new package just to avoid am unliked built-in feature of ES6 is justified :/

wtrocki:
Got that running and checked the code. 
@josemigallas Could you take look on this as well.

ui/components/common/CommonToolbar.js
wojta: ok, then :/

wtrocki:
Awesome work Josemi

PR #117, opened 2018-08-27
WIP: Update resolvers after datasource is deleted

psturc:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7824
## Checklist:

- [x] Changes have been successfully verified by another team member 

## Progress

- [x] Refetching resolvers so deleted ones no longer appear in the list of resolvers
- [x] If deleted resolver was being edited before deleting, return resolver editor to initial state


ui/components/data-sources/DeleteDataSourceDialog.js
```diff
@@ -21,16 +22,26 @@ const DeleteDataSourceDialog = ({
...
+        const queriesToRefetch = [{
+            query: GetDataSources,
+            variables: { name }
+        }, {
+            query: GetDataSources,
+            variables: { name: undefined }
+        }];
+
+        if (hasResolvers) {
+            dataSource.resolvers.forEach(resolver => {
```
josemigallas: Use `map` instead.

ui/components/data-sources/DeleteDataSourceDialog.js
```diff
@@ -21,16 +22,26 @@ const DeleteDataSourceDialog = ({
...
+                queriesToRefetch.push({
```
josemigallas: I think this is a bit overkilling, considering the schemaId and type will be the same in many cases. I wonder if it would be possible to somehow simplify the refetch, something like
```
{
    query: GetResolvers,
    variables: { all: true }
}
```

josemigallas: Or:
```
dataSource.resolvers.forEach(resolver => {
    const query = {
        query: GetResolvers,
        variables: { schemaId: resolver.GraphQLSchemaId, type: resolver.type }
    };
    if (!queriesToRefetch.indexOf(query)) {
        queriesToRefetch.push(query);
    }
});
```

PR #125, opened 2018-09-05
Prevent selecting same resolver twice

josemigallas:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7834

## What
Checks if clicked resolver is the one already selected and if so do nothing

## Why
UX improvement

## Verification Steps

1. Go to Resolvers, select any
2. Then do any change, so that the Save button is enabled
3. Without saving it, select the same resolver again: nothing should happen
4. Without saving it, select a different resolver: a warning should pop up

## Checklist:

- [x] Changes have been successfully verified by another team member

### DIFF
--- a/ui/components/resolvers/ResolversContainer.js
+++ b/ui/components/resolvers/ResolversContainer.js
@@ -25,13 +25,19 @@ class ResolversContainer extends Component {
         return [];
     }
 
-    onResolverClicked(resolver) {
-        const { isResolverSaved } = this.state;
+    onResolverClicked(clickedResolver) {
+        const { resolver, isResolverSaved } = this.state;
+
+        if (resolver !== null
+            && resolver.type === clickedResolver.type
+            && resolver.field === clickedResolver.field) {
+            return;
+        }
 
         if (isResolverSaved) {
-            this.setState({ resolver });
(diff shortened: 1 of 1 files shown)


ui/components/resolvers/ResolversContainer.js
```diff
@@ -25,13 +25,19 @@ class ResolversContainer extends Component {
...
+            && resolver.field === clickedResolver.field) {
+            return;
+        }
 
         if (isResolverSaved) {
-            this.setState({ resolver });
+            this.setState({ resolver: clickedResolver });
         } else {
-            this.setState({ showChangeConfirmation: true, clickedResolver: resolver });
+            this.setState({ showChangeConfirmation: true, clickedResolver });
```
pb82: So the state has two properties for the resolver, `resolver` and `clickedResolver`. Is that intentional?

josemigallas: That is indeed intentional. `clickedResolver` is a temporal variable while the confirmation dialog is opened. In case the user cancels, it gets deleted but if they confirm, then `resolver = clickedResolver`.
In other words, `this.state.resolver` is the one controlling `ResolverDetail`.

PR #132, opened 2018-09-11
adds some testing info to the readme

josemigallas:
Adds some instructions and recommendations for running and adding tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,3 +70,44 @@ A running PostgreSQL server. To run postgres in a docker container use the follo
 UI server has some environment variables to be set. If they're not set, defaults for development will be used.
 
 * `AUDIT_LOGGING`:   : If true, audit logs of resolver operations will be logged to stdout. Defaults to true.
+ 
+# Tests
+
+The UI makes use of [Jest](https://jestjs.io/) as assertion library and [Enzyme](http://airbnb.io/enzyme/) as DOM rendering and manipulation utility.
+
+## Running the test suite
+To execute the whole test suite:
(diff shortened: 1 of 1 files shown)


matskiv:
LGTM, but if you could also explain a bit how this workaround works and why we need it, it would be perfect - `await sleep(0); // Wait for the query to finish`.

josemigallas:
@matskiv done! Feel free to merge.

README.md
```diff
@@ -69,4 +69,45 @@ A running PostgreSQL server. To run postgres in a docker container use the follo
...
+
+* Test files must end with `.test.js` for Jest to find them.
+* Components using GraphQL queries must be fully rendered and inside a `MockedProvided` component:
+   ```javascript
+    <MockedProvider mocks={[]} addTypename={false}>
+        <DataSourcesList />
+    </MockedProvider>
+   ```
+   Read more about testing with Apollo and React ah the [Apollo testing guide](https://www.apollographql.com/docs/guides/testing-react-components.html).
+* When asserting using query result, you must define mocks for all involved ones. They have to be mocked in the exact way they are called and they are only valid for 1 call. Moreover, this process is async and the compoent first render won't have any results yet (potentially ending up with a false negative). To workaround this, use the utility function `sleep()` and `wrapper.update()` in order for the queries to finish. This [Github issues answer](https://github.com/apollographql/react-apollo/issues/1711#issuecomment-369511476) describes briefly and concisely how to do it.
```
matskiv: @josemigallas  small typo - "compoent"

PR #133, opened 2018-09-14
Investigation - Resolver generator

wtrocki:
## Motivation

WIP. Posting for visibility.
File changes do not matter (done this hacky way for the moment). 
## How to try this out

```
git checkout generator-wip
npm i
npm run build server
```

@pb82 There are two spikes here. 

## Creator Spike:

One is building generic function based resolvers and attaching them into the schema. Same happens with schema. For this use case there is no way to edit additional schema or resolvers as they do not follow request/response pattern. This is not using our underlying execution framework, thus there is no way to see schema or resolvers, but introspection on the right will still show everything - we could then add some actions on this generated elements etc.

## Generator Spike:

Second one that is presented on gif is where we using our underlying server execution mechanism and physically generate new objects in schema and resolvers.  Users see and can edit both schema and resolvers. In this aproach users get instant flexibility and we can apply schema first aproach.

Please do not mind UI code - this is only for spike purposes and going to be refactored later.
Code is based on https://github.com/wtrocki/apollo-resolver-gen

Diagrams that shows what I'm talking about here

## Approach 1 

screen shot 2018-09-18 at 5 18 33 pm

## Approach 2 
screen shot 2018-09-18 at 5 18 43 pm


pb82:
@wtrocki Ok, thanks for the clarification. I was just wondering because you showed the generated resolver in the UI Resolver Editor.

wtrocki:
Outdated. Closing

PR #135, opened 2018-09-18
add ability for UI to be standalone

StephenCoady:
# Motivaton

This change will allow us to address some technical debt and architecture limitations.
Most important ones:
- Add ability to try out queries and schemas without publishing them to the server
- Add ability to Query schema when the server is protected
- Separate UI from Server (server will be production environment and UI will be development environment)
- Add support for additional use cases (user provided sync server can use our execution later)
- Improve and fix schema reload for sync
- Unify models that are being used in UI and Server (so they will be extended in one single place)
- Support schema seeding (memeo and other schemas - this is currently happening on Server - we need that ability on UI)

TL;DR - This is a crucial change that fixes some problems with the Sync architecture and it will give us tons of benefits with very little investment.

# Side effects of this change

- UI and Server will reuse the same module
- Development time will require using github version of new module (or releases to npm)
When we will get more and more modules we can consider moving towards mono repositories etc. but currently this impact is marginal

# Verification Steps

To verify that these changes work, it can be done as part of the overall verification of the feature. Using this repository and the sync UI.

1. Use docker-compose to create the db locally. 
1. Run the server with `npm run dev:keymemeo`
1. Run the UI with `npm run server`
1. Everything should work as expected with the server and the UI. i.e. authentication should be required on the server and not on the UI.


wtrocki:
# Motivaton

This change will allow us to address some technical debt and architecture limitations.
Most important ones:
- Add ability to try out queries and schemas without publishing them to serevr
- Add ability to Query schema when server is protected
- Closely separate UI from Server (server will be production environment and UI will be development environment)
- Add support for additional use case (user provided sync server can use our execution later)
- Improve and fix schema reload for sync
- Unify models that are being used in UI and Server (so they will be extended in one single place)
- Support schema seeding (memeo and other schemas - this is currently happening on Server - we need that ability on UI)

TL;DR - This is the crucial change that fixes some problems with Sync architecture and it will give us tons of benefits with very little investments.

# Side effects of this change

- UI and Server will reuse the same module
- Development time will require using github version of new module (or releases to npm)
When we will get more and more modules we can consider moving towards mono repositories etc. but currently this impact is marginal

server/playground/index.js
```diff
@@ -0,0 +1,28 @@
+const buildSchema = require("data-sync-gql-core");
```
wtrocki: Since this will be the first npm package for sync we may consider using `@aerogear` org for it. `@aerogear/data-sync-gql-core`

@darahayes Opinions?

server/playground/index.js
```diff
@@ -0,0 +1,28 @@
...
+const { makeExecutableSchema } = require("graphql-tools");
+
+const { ApolloServer } = require("apollo-server-express");
+const models = require("../models");
+
+function connectDataSources(dataSources) {
```
wtrocki: We can refactor top level api to offer something like this:

interface Core {
 buildSchema(schemaName: string): Promise<Schema>
 createGraphQLApi(config: SyncCoreConfig) : Promise<void>  // { app, path: "/graphqldata" }
 connectActiveDataSources(): Promise<void>

ui/components/queries/QueriesContainer.js
```diff
@@ -9,7 +9,7 @@ const settings = {
 
 const QueriesContainer = () => (
     <Provider store={store}>
-        <Playground endpoint="http://localhost:8000/graphql" settings={settings} />
+        <Playground endpoint="/graphqldata" settings={settings} />
```
wtrocki: [Note] We may have build time parameter for this.

server/playground/index.js
darahayes: I'm +1 on using `@aerogear`

server/playground/index.js
```diff
@@ -0,0 +1,28 @@
...
+    Object.keys(dataSources).forEach(key => {
+        const dataSource = dataSources[key];
+        dataSource.connect();
+    });
+}
+
+
+module.exports = app => {
```
darahayes: Minor thing: This function is referenced from the `server/server.js` as `ApplyExecutableMiddleware`. It that's the name we're using there, then we should name this function appropriately here.

Additionally, this function is performing some async operations

* Building the schema (see the `.then`)
* Connecting data sources.

However, the function is made to look like it's synchronous and it does not surface any errors that might happen during either of those operations. So in `server.js` when we call `ApplyExecutableMiddleware(App);` we have no idea if that was successful or not.

The function should either use `async/await` (more preferable) or return a promise.

server/server.js
```diff
@@ -16,10 +18,12 @@ App.use(json());
 App.use(express.static(join(__dirname, "../public")));
 App.use(expressPino);
 
-App.use("/", router);
 
 let server = null;
 
+App.use("/", serverApi());
+ApplyExecutableMiddleware(App);
```
darahayes: Like I mentioned, this should be using a promise or `async/await`.

So either 

```
ApplyExecutableMiddleware(App).then(() => {
  //success
}).catch((error) => {
  // oops something happened - perhaps we should crash
})
```

or 

```
try {
  await ApplyExecutableMiddleware(App)
} catch(error) {
  // oops
}
```

Note that it is not possible to use `async/await` in a top level script though, it has to be inside an `async function` so perhaps this file could be refactored to run inside a function.

darahayes (commented):
@StephenCoady Overall, this PR shows off how valuable the work around splitting out the execution layer has been. Look how little code there is to make this work! It's amazing.

I know this is a work in progress so you're probably planning on tidying up the code later but I left some small feedback that might be useful when you go to do the tidy up

StephenCoady:
@wtrocki and @darahayes thanks for the feedback. Will be making these changes - thanks.

StephenCoady:
@darahayes @wtrocki mind taking another look? all changes addressed and the core module updated to have sync history. thanks.

server/models/index.js
```diff
@@ -51,9 +51,13 @@ const reset = () => database.dropAllSchemas({ logging: false });
 module.exports = {
     database,
     dataSource,
+    DataSource: dataSource,
```
wtrocki: I was hoping that we are going to unify that. It should not be huge deal.

server/playground/index.js
```diff
@@ -0,0 +1,33 @@
+const buildSchema = require("@aerogear/data-sync-gql-core");
+const { makeExecutableSchema } = require("graphql-tools");
+const { ApolloServer } = require("apollo-server-express");
+const { log } = require("../logger");
+const models = require("../models");
+
+async function connectDataSources(dataSources) {
+    log.info("Connecting data sources");
+    Object.keys(dataSources).forEach(async key => {
```
wtrocki: I think nested async functions do not work great and are source of bugs.
Could we replace that with `for of` loop that do not require function?

server/playground/index.js
```diff
@@ -0,0 +1,33 @@
+const buildSchema = require("@aerogear/data-sync-gql-core");
+const { makeExecutableSchema } = require("graphql-tools");
+const { ApolloServer } = require("apollo-server-express");
+const { log } = require("../logger");
+const models = require("../models");
+
+async function connectDataSources(dataSources) {
```
wtrocki: In my mind this should be responsibility of the core module (however I do not have strong opinion about it). See my previous proposition in comment for end user api.

server/playground/index.js
```diff
@@ -0,0 +1,33 @@
...
+            throw (error);
+        }
+    });
+}
+
+
+async function ApplyExecutableMiddleware(app) {
+    // read default schema
+    // Attach middleware to app
+    const { schema, dataSources } = await buildSchema(models, null, null, makeExecutableSchema);
```
wtrocki: My assumption is that this is going to read "default" schema. 
We could allow users to pass string so the schema will be configurable in core?
Could models be embedded into core (they are there already anyway) so end user api will be like: `createMiddlewareFor({schema:"default"})`

This aproach is working but TBH merging it this way will introduce tech debt.

ui/components/playground/PlaygroundContainer.js
```diff
@@ -9,7 +9,7 @@ const settings = {
 
 const PlaygroundContainer = () => (
     <Provider store={store}>
-        <Playground endpoint="http://localhost:8000/graphql" settings={settings} />
+        <Playground endpoint="/graphqldata" settings={settings} />
```
wtrocki: [Optional] Ideally this should be done by fetching graphql query (currently it could return some fixed string).

server/server.js
```diff
@@ -2,8 +2,8 @@ const { urlencoded, json } = require("body-parser");
 const express = require("express");
 const { join } = require("path");
 const { log, expressPino } = require("./logger");
-const { port } = require("./config");
-const { sync, database } = require("./models");
+const { port, postgresConfig } = require("./config");
+const { sequelize } = require("@aerogear/data-sync-gql-core").models(postgresConfig);
```
wtrocki: This is quite complex import. Let's move that outside import section.

server/routes/index.js
```diff
@@ -5,7 +5,9 @@ const express = require("express");
 const { runHealthChecks } = require("../health");
 const { log } = require("../logger");
 const { compileSchemaString } = require("../gql/helper");
-const { schema } = require("../models");
+const { postgresConfig } = require("../config");
+const { GraphQLSchema } = require("@aerogear/data-sync-gql-core").models(postgresConfig);
```
wtrocki: Why we do not use getModels from api? This seems to be initializing models for every file. 
We could create single module here that is doing that once or setup core based on postgresConfig.

server/playground/index.js
```diff
@@ -1,33 +1,16 @@
...
-    // Attach middleware to app
-    const { schema, dataSources } = await buildSchema(models, null, null, makeExecutableSchema);
+    const core = new Core(config, makeExecutableSchema);
+    const { schema, dataSources } = await core.buildSchema("default");
     const apolloServer = new ApolloServer({
         schema
     });
     apolloServer.applyMiddleware({ app, path: "/graphqldata" });
-    await connectDataSources(dataSources);
+    await core.connectActiveDataSources(dataSources);
```
wtrocki: No strong opinion here - more asking question: Do you think we should wait for all datasources to connect before we will enable middleware?

server/playground/index.js
```diff
@@ -1,33 +1,16 @@
...
-            log.error(error);
-            throw (error);
-        }
-    });
-}
-
+const { Core } = require("@aerogear/data-sync-gql-core");
+const config = require("../config").postgresConfig;
 
 async function ApplyExecutableMiddleware(app) {
```
wtrocki: first letter could be small as this is an function not an class (not your change, but let's improve it)

wtrocki:
Newbie question: Should we remove models from: https://github.com/aerogear/data-sync-ui/tree/master/server/models

server/gql/models.js
```diff
@@ -1,6 +1,11 @@
 const core = require("@aerogear/data-sync-gql-core");
-const { postgresConfig } = require("../config");
+const { postgresConfig, sqLiteConfig } = require("../config");
 
-const models = core.models(postgresConfig);
+let models = null;
```
wtrocki: Avoid null (undefined is ok)

StephenCoady:
@wtrocki I think this is ready to do a final verification on now.

server/server.js
```diff
@@ -3,9 +3,11 @@ const express = require("express");
 const { join } = require("path");
 const { log, expressPino } = require("./logger");
 const { port } = require("./config");
-const { sync, database } = require("./models");
+const { sequelize } = require("./gql/models");
 const { stopNotifications } = require("./configNotifiers/configNotifierCreator");
-const router = require("./routes");
+const serverApi = require("./routes");
+
+const applyExecutableMiddleware = require("./playground");
```
darahayes: At first glance, this line of code has very little meaning to me. I'm wondering why is there a file called `playground` and why does it export a function called `applyExecutableMiddleware`, and what on earth is an executable middleware?

If we leave this function named as `applyExecutableMiddleware` then I fear the term 'executable middleware' will be casually thrown around in conversation all the time when referring to a specific instance of the core execution layer (especially when we start dealing with multiple instances at the same time). To me that term makes no sense and would be confusing to someone without the proper context.

I would suggest we find a better name for the file and for the function within.

I know it's a mouthful but perhaps something like `applyExecutionLayer` or `applyExecutionLayerForSchema` (it is an execution layer for a specific schema after all). With this kind of naming, a newcomer would be asking _what's an execution layer?_ and that's a good question to ask.

darahayes (changes requested):
@StephenCoady left some feedback! Great work overall.

StephenCoady:
@darahayes mind taking another look? I have added a query to fetch a (hardcoded) endpoint for now. I also changed the name of the playground folder and its function based on your comments - they make much more sense now.

ui/components/playground/PlaygroundContainer.js
```diff
@@ -1,16 +1,30 @@
...
-    </Provider>
+    <Query query={GetGraphQLEndpoint}>
+        {({ loading, error, data }) => {
+            if (loading || typeof error !== "undefined") {
+                return <FormControl.Static>Loading graphql playground...</FormControl.Static>;
+            }
+            const { getGraphQLEndpoint } = data;
+            return (
+                <Provider store={store}>
+                    <Playground endpoint={getGraphQLEndpoint} settings={settings} />
```
wtrocki: 😍

wtrocki: @darahayes Why we need to do this now? Are we trying to build project functionality? 
This has too big scope already.  PR and work is not designed to work with the projects abstractions and as you mentioned it will be trivial to implement, however let's not add more to this work and focus on the targets of this sprint. 

TL;DR - with this changes projects abstraction will work, but let's not add it here.
I was hoping to reduce scope of this changes and leave some obvious places where we can hook on later.

wtrocki: Actually if that is small change then it's ok. I would to this exactly as suggested. 
This PR seems awesome.

darahayes (approved):

darahayes:
Verified. I was able to make queries without a running data sync server. Amazing work @StephenCoady!

PR #137, opened 2018-10-04
Adding mongo data source type

wtrocki:
## Motivation

Investigation: Add mongodb datasource.

Required to run: 
https://github.com/aerogear/data-sync-server/pull/100

### DIFF
--- a/server/gql/schema.js
+++ b/server/gql/schema.js
@@ -8,7 +8,8 @@ const dataSourceValidator = require("../dataSourceValidator/dataSourceValidator"
 const Schema = buildSchema(`
     enum DataSourceType {
         InMemory,
-        Postgres
+        Postgres,
+        Mongo
     },
     type Query {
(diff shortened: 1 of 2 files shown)


wtrocki:
See https://github.com/aerogear/data-sync-server/pull/100

pb82 (approved):
This is a good start. To actually implement the data source configuration you need to add the option here: https://github.com/aerogear/data-sync-ui/blob/master/ui/components/data-sources/BaseDataSourceDialog.js#L263

wtrocki:
@pb82 I'm trying to prove that we could use mongo for both metadata and data so users could choose the right db at the start. This is not going to be merged soon - it's more investigation how to properly model interfaces for data so we will not be tied to single datasource. Got really nice results so we should be able to move forward with actual requirements.

wtrocki:
Outdated.

